PR #6, opened 2015-02-05
Use tokenizer for finding quotes.

2m:
I know that this revamps completely how quotes are found, but using tokenizer allows catching such violations as:

* double quotes in multiline string where double quote appears not at the same line as start or the stop of the string

Also added tests.

This does not work on Python 2. If you want Python 2 support, let me know, I will add it.


2m:
This PR now also has Python 2 support thanks to @kvdb who submitted a PR against this PR (https://github.com/2m/flake8-quotes/pull/1).

I have also added CI setup using `tox` which now runs tests under Python 2.7 and 3.4. [This is](https://travis-ci.org/2m/flake8-quotes/builds/52944162) how it looks like on my fork.

@zheller you can enable CI on this repo by flicking a switch [here](https://travis-ci.org/profile/zheller).

Also converted readme from markdown to rst as otherwise we would require to create `MANIFEST.in` file to bring in markdown readme to source distribution.

PR #11, opened 2015-10-30
Added release script

twolfson:
Releasing can be tedious and error-prone with humans. As a result, I prefer to automate via scripts. I typically use:

https://github.com/twolfson/foundry

Unfortunately, my Python library doesn't currently support non-`setup.py` based versions and I am heads down on other projects. As a result, I am going with a quick and dirty `release.sh` script for now.

/cc @zheller

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+# Exit on first error
+set -e
+
+# Parse our CLI arguments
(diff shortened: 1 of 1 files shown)


zheller:
Shipit

twolfson:
Shipped and released as `0.1.1` :+1:

PR #16, opened 2016-02-12
Use pep8 input functions to avoid stdin issues

zedlander:
Inspired by [some code in flake8-import-order](https://github.com/public/flake8-import-order/blob/master/flake8_import_order/__init__.py#L208).

Runs correctly for me in python2.7 and python3.5 alongside both flake8-import-order and flake8-commas. Can't guarantee it plays nicely with other plugins, but it does work where sys.stdin.read() fails, so pep8 is at least doing something smart.

### DIFF
--- a/flake8_quotes.py
+++ b/flake8_quotes.py
@@ -40,7 +40,6 @@ class QuoteChecker(object):
 
     def get_file_contents(self):
         if self.filename in ('stdin', '-', None):
-            self.filename = 'stdin'
             return pep8.stdin_get_value().splitlines(True)
         else:
             return pep8.readlines(self.filename)
--- a/flake8_quotes.py
+++ b/flake8_quotes.py
@@ -1,6 +1,6 @@
-from sys import stdin
 import tokenize
 
+import pep8
 
 __version__ = '0.1.2'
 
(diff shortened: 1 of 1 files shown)


flake8_quotes.py
```diff
@@ -39,14 +39,14 @@ def parse_options(cls, options):
         cls.quotes = cls.QUOTES[options.quotes]
 
     def get_file_contents(self):
-        with open(self.file, 'r') as file_to_check:
-            return file_to_check.readlines()
+        if self.filename in ('stdin', '-', None):
+            self.filename = 'stdin'
```
twolfson: Do we need to set `self.filename` here? I feel like we are contaminating the source variable =/

twolfson:
Looks good for the most part, left a comment on 1 line

zedlander:
That came from flake8-import-order, but they probably have more logic looking at the filename. Nuked it.

twolfson:
At a glance everything looks good, I will prob want to play around with this on the CLI before landing. I do that by the end of this weekend.

Thanks for the great PR and quick responses =)

zedlander:
:+1: thanks!

twolfson:
Taking a look at this now

twolfson:
Everything seems to work, I have no idea what was going on in my older branch (it seemed to work as well). This has been landed/released in `0.2.0`. Thanks for the PR!

PR #18, opened 2016-02-12
Update setup.py

do3cc:
Currently, 0.2.1 is broken for us, because this dependency is not declared and does not get installed.
Also, it would be nice if you added a Changelog file.


twolfson:
Ah, strange. I thought the `flake8` install would catch it.

twolfson:
This has been merged/released in `0.2.2`. Thanks for the fix!

do3cc:
thanks!

PR #20, opened 2016-02-12
Move version number to separate file

### LINKED ISSUE #19: Error: Couldn't install: flake8-quotes 0.2.2
```bash
$ ./bin/buildout 
Develop: '/home/user/sites/brasil.gov.integracaocaldav/.'
Unused options for buildout: 'package-min-coverage'.
Updating instance.
Updating test.
Installing code-analysis.
Getting distribution for 'flake8-quotes'.
  File "<string>", line 1, in <module>
  File "/home/user/cache-python/eggs/setuptools-0.6c11-py2.7.egg/setuptools/command/easy_install.py", line 1712, in main
  File "/home/user/cache-python/eggs/setuptools-0.6c11-py2.7.egg/setuptools/command/easy_install.py", line 1700, in with_ei_usage
  File "/home/user/cache-python/eggs/setuptools-0.6c11-py2.7.egg/setuptools/command/easy_install.py", line 1716, in <lambda>
  File "/home/user/pythons/python279/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/home/user/pythons/python279/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/home/user/pythons/python279/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/home/user/cache-python/eggs/setuptools-0.6c11-py2.7.egg/setuptools/command/easy_install.py", line 211, in run
  File "/home/user/cache-python/eggs/setuptools-0.6c11-py2.7.egg/setuptools/command/easy_install.py", line 427, in easy_install
  File "/home/user/cache-python/eggs/setuptools-0.6c11-py2.7.egg/setuptools/command/easy_install.py", line 476, in install_item
  File "/home/user/cache-python/eggs/setuptools-0.6c11-py2.7.egg/setuptools/command/easy_install.py", line 655, in install_eggs
  File "/home/user/cache-python/eggs/setuptools-0.6c11-py2.7.egg/setuptools/command/easy_install.py", line 930, in build_and_install
  File "/home/user/cache-python/eggs/setuptools-0.6c11-py2.7.egg/setuptools/command/easy_install.py", line 919, in run_setup
  File "/home/user/cache-python/eggs/setuptools-0.6c11-py2.7.egg/setuptools/sandbox.py", line 62, in run_setup
  File "/home/user/cache-python/eggs/setuptools-0.6c11-py2.7.egg/setuptools/sandbox.py", line 105, in run
  File "/home/user/cache-python/eggs/setuptools-0.6c11-py2.7.egg/setuptools/sandbox.py", line 64, in <lambda>
  File "setup.py", line 3, in <module>
    from setuptools import find_packages
  File "/tmp/easy_install-7uVzSE/flake8-quotes-0.2.2/flake8_quotes.py", line 3, in <module>
ImportError: No module named pep8
An error occurred when trying to install flake8-quotes 0.2.2. Look above this message for any errors that were output by easy_install.
While:
  Installing code-analysis.
  Getting distribution for 'flake8-quotes'.
Error: Couldn't install: flake8-quotes 0.2.2
```

Ichimonji10:
flake8-quotes cannot be installed due to a cyclical dependency: `setup.py` imports module `flake8_quotes` in order to get at its version number, but that module cannot be imported until flake8-quotes has been installed.

Move the version number to an independent file. This file is then read as needed. Update version number references as appropriate.

As a nitpick, upper-case all constants in module `setup.py`.

Given the following environment for both `N=2` and `N=3`:

```sh
virtualenv --python python$N ~/.virtualenvs/flake8-quotes$N
source ~/.virtualenvs/flake8-quotes$N/bin/activate
```

The following commands succeed:

```sh
pip install -e .
python setup.py develop
python setup.py install
```

Fix #19.


idgserpro:
:+1:

flake8_quotes.py
```diff
@@ -1,8 +1,11 @@
+import os
 import tokenize
 
 import pep8
 
-__version__ = '0.2.2'
+
+with open(os.path.join(os.path.dirname(__file__), 'VERSION')) as handle:
```
twolfson: We should use an `r` flag for our open to be explicit

twolfson:
This seems like it should work. Nice job. Left a comment to fix about `open` though

flake8_quotes.py
Ichimonji10: Sure. Will do.

Ichimonji10: Doing this overrides the default of `'rt'`, which may be preferable in this case:

> Python distinguishes between files opened in binary and text modes, even when
> the underlying operating system doesn't. Files opened in binary mode
> (appending 'b' to the mode argument) return contents as bytes objects without
> any decoding. In text mode (the default, or when 't' is appended to the mode
> argument), the contents of the file are returned as strings, the bytes having
> been first decoded using a platform-dependent encoding or using the specified
> encoding if given.

Do you still want to explicitly pass `'r'` here? We're working with a text file, and it's probably better to have the interpreter explicitly perform decoding for us, instead of implicitly trusting Python to do the right then when asked to `strip()` some bytes.

twolfson: `r` should be fine; there is no special encoding being used on `VERSION` and we own it. I don't think `rt` is Python 2 compatible.

Ichimonji10: Good enough.

```python
>>> b'foo bar   '.strip()
b'foo bar'
```

Ichimonji10: Thanks for the pointer re: Python 2 vs 3 compatibility.

Ichimonji10:
Updated.

twolfson:
Awesome, still on track to try out/review this by midnight CST tonight

twolfson:
Tried it out via the commands below. Looks good to me, landing this release. Thanks for your work on this :+1: 

```bash
# Reset environment
rmvirtualenv flake8-quotes
mkvirtualenv flake8-quotes
git checkout master

# Reproduce error via `pip install`
pip install -e .
# Command "python setup.py egg_info" failed with error code 1 in /home/todd/github/flake8-quotes

# Check out patch PR
git fetch pr # Added via `git add-pr` from my `dotfiles`
git checkout pr/20

# Verify patch works
pip install -e .

# Verify flake8 still works
pip install flake8
flake8 .
# ./test/data/doubles.py:1:25: Q000 Remove bad quotes.
```

twolfson:
This has been merged/released in `0.2.3`.

Ichimonji10:
Great! Thank you, both for the new release and for the plugin in general. (It lets me about much bikeshedding.)

PR #21, opened 2016-02-15
Fetch package version at runtime w/setuptools

Ichimonji10:
The `VERSION` file contains the flake8_quotes version number. This file is
included with the project when its repository is cloned by virtue of being under
source control, and it is included in installable packages by virtue of being
mentioned in `MANIFEST.in`. However, this file is never actually installed. Only
files within the `flake8_quotes` namespace are available after installation.

This is problematic because the `flake8_quotes` module includes a reference to
this file. Installing flake8_quotes and then importing said module will trigger
an import error, as the `VERSION` file is unavailable after installation. Fix
this issue by doing the following to module `flake8_quotes`:

1. Drop the reference to the `VERSION` file.
2. Use setuptools to load the version number (from `metadata.json`).

### DIFF
--- a/flake8_quotes.py
+++ b/flake8_quotes.py
@@ -1,11 +1,10 @@
-import os
 import tokenize
 
 import pep8
+import pkg_resources
 
 
-with open(os.path.join(os.path.dirname(__file__), 'VERSION'), 'r') as handle:
-    __version__ = handle.read().strip()
+__version__ = pkg_resources.get_distribution('flake8_quotes').version
 
 
 class QuoteChecker(object):


Ichimonji10:
Sorry about this mishap. I really should've more extensively tested #20. This time around, I tested things by cloning a different project that depends on flake8-quotes and installing the updated version of this project like so:

```sh
virtualenv env
source env/bin/activate
pip install --no-cache-dir -r requirements.txt -r requirements-dev.txt
make lint-flake8  # triggers call to flake8-quotes
```

In the above case, `requirements-dev.txt` included the following line:

    file:///home/ichimonji10/code/flake8-quotes#egg=flake8-quotes

Ichimonji10:
Of possible interest:

* https://packaging.python.org/en/latest/single_source_version/
* https://caremad.io/2013/07/setup-vs-requirement/

twolfson:
Can you explain why the `VERSION` file is unavailable after installation? We are including it in the `MANIFEST.in` =/

Ichimonji10:
tl;dr: Only files in the `flake8_quotes` namespace are available after installation. In our case, that means `flake8_quotes.py`.

`python setup.py sdist` or `python setup.py bdist_wheel --universal` [1] instructs Python to create an installable egg/wheel. What's included in this egg/wheel? It includes whatever files are named by the auto-generated `MANIFEST` file, plus whatever is named by the `MANIFEST.in` file.

But we're not done yet. A user doesn't run flake8-quotes directly from one of these eggs/wheels. [2] Instead, they must install an egg/wheel, and *then* they can use flake8-quotes. And not everything in an egg/wheel is available after that egg/wheel is installed.

This actually makes a lot of sense. Let's say you put unit tests in a top-level `tests` directory. Do unit tests need to be importable and available for use after installation? No, of course not. What about `release.sh`, `README.rst`, `setup.py`, and so on? They don't need to be available for use after installation either. And so they aren't. You can prove this to yourself by creating a virtualenv, installing a package into it (with `python setup.py install`, not `python setup.py devel`), and inspecting it. Try this:

```sh
virtualenv env
source env/bin/activate
pip install --no-cache ~/code/flake8-quotes
# or cd ~/code/flake8-quotes && python setup.py install && cd -
tree env
```

Here's a snippet from my system:

```
│       ├── site-packages
│       │   ├── easy_install.py
│       │   ├── flake8_quotes-0.2.3-py3.5.egg-info
│       │   │   ├── dependency_links.txt
│       │   │   ├── entry_points.txt
│       │   │   ├── installed-files.txt
│       │   │   ├── PKG-INFO
│       │   │   ├── requires.txt
│       │   │   ├── SOURCES.txt
│       │   │   ├── top_level.txt
│       │   │   └── zip-safe
│       │   ├── flake8_quotes.py
```

[1] When creating a wheel, you can also execute `python setup.py bdist_wheel`. However, adding the `--universal` flag is recommended if distributing a pure-Python package.
[2] There is a technique for running code directly from `.zip` archives without installing it, but that's beyond the scope of this discussion.

twolfson:
Can you link me to the reference material for all of this?

Ichimonji10:
Sure. I think the [Python Packaging User Guide](https://packaging.python.org/en/latest/) is a great guide. Its [packaging and distributing projects](https://packaging.python.org/en/latest/distributing/) section is most relevant to this discussion. In addition to that, the following is also relevant:

* [creating a source distribution](https://docs.python.org/3.5/distutils/sourcedist.html) discusses the manifest files.
* [adding non-code files](http://python-packaging.readthedocs.org/en/latest/non-code-files.html), from [how to package your python code](http://python-packaging.readthedocs.org/en/latest/index.html) discusses techniques for installing additional files (like a `VERSION` file).
* [writing the setup script](https://docs.python.org/3/distutils/setupscript.html) is the One True Reference for this topic. It's also the most in-depth.

As alluded to, there are other ways of solving the issue tackled by this pull request. Passing `include_package_data=True` to `setup()` is one approach; the `package_data` and `data_files` arguments may also be suitable. Given that these options provide the prospect of including the VERSION file, they may actually be preferable.

twolfson:
Cool, thanks for the info. My hesitation with the current approach is `pkg_resources` is not in Python's core and will likely cause issues for people who use non-standard install methods (e.g. `.deb` packages).

I think I would prefer:

- Using `include_package_data=True`
- Using an approach like https://github.com/public/flake8-import-order/tree/2ac7052a4e02b4a8a0125a106d87465a3b9fd688/flake8_import_order
    - Store `__version__` in an `__about__.py` and `exec` that inside of `setup.py`
    - https://github.com/public/flake8-import-order/blob/2ac7052a4e02b4a8a0125a106d87465a3b9fd688/flake8_import_order/__about__.py
    - https://github.com/public/flake8-import-order/blob/2ac7052a4e02b4a8a0125a106d87465a3b9fd688/setup.py#L8-L9

> I mentioned being unsure about `flake8` needing `version` but it definitely does as we set `version` as a class property. https://flake8.readthedocs.org/en/latest/extensions.html

Ichimonji10:
Given that `install_requires=['setuptools', ...]`, `pkg_resources` will always be available. (Any competent third-party packager should respect the `install_requires` argument.)

twolfson:
Yea, it makes sense that it should work in most cases but when I search for "pkg_resources" in Google, I see a lot of issues. As a result, I would like to avoid that path.

Ichimonji10:
It's natural that a google search for "pkg_resources" will turn up many issues. It's one of the most popular Python packages in existence, with 6.6 million installs in the last month. Any minor mistake made by setuptools itself, or by any Python package that uses setuptools, or by any packager that uses setuptools, is going to affect many people.

Again, it's totally fine to handle version numbers some other way, or to choose not to expand test coverage for an application. There are valid reasons for doing so. But please, take the time to find them, or at least state that the choices made are more or less arbitrary.

twolfson:
@Ichimonji10 Are you planning on opening another PR that handles the suggested options from https://github.com/zheller/flake8-quotes/pull/21#issuecomment-184383442 ?

Ichimonji10:
@twolfson No, not at this point.

twolfson:
Okay, thanks for the quick reply

do3cc:
@twolfson Just fyi regarding not including setuptools, it is so widespread that pip uses it by default in some cases instead of distutils: https://packaging.python.org/en/latest/current/#id14

PR #22, opened 2016-02-15
Update supported Python versions

Ichimonji10:
Currently, Travis tests flake8_quotes under Python 3.3, and `setup.py` declares
flake8_quotes as being available for Python 3.4. Address this discrepancy by
making Travis test flake8_quotes under Python 3.3 – Python 3.5, and listing the
same set of versions in `setup.py`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ language: python
 python:
   - "2.7"
   - "3.3"
+  - "3.4"
+  - "3.5"
 
 install:
   # Install our dependencies
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,9 @@ setup(
         'Programming Language :: Python :: 2',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.3',
         'Programming Language :: Python :: 3.4',
+        'Programming Language :: Python :: 3.5',
         'Topic :: Software Development :: Libraries :: Python Modules',
         'Topic :: Software Development :: Quality Assurance',
     ]


twolfson:
Can we only test with one version of Python 3? While I know that we can test against every version, it's often unnecessary and adds undesirable build time to developers.

Ichimonji10:
Sure, you can. But that gives you less assurance that flake8-docstrings works well with the advertised versions of Python. I find that useful, as I personally run Python 2.7 and 3.5, and work with systems that run 3.4.

If you're worried about build times, Travis offers [container-based builds](https://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure), which are *very* fast. It requires adding `sudo: false` to `.travis.yml`. As for devs who use tox, well, I can't offer any comfort.

twolfson:
I understand the trade-offs but but would still prefer to only test 1 verison of Python 3; there is nothing in this repo that is likely to break across minor versions.

Ichimonji10:
I'd really like to press you about the "adds undesirable build time to developers" bit. Is that true? The tox configuration is not being touched, I don't think. And I can personally vouch for how fast Travis's container-based builds are. Basically: what costs are being avoided here?

twolfson:
It's the cost of maintenance for something that doesn't need to be maintained

Ichimonji10:
Ahh, so testing this project across multiple Python 3 versions doesn't add appreciable build time. Instead, this pull request raises the worrisome prospect of additional maintenance effort.

That's understandable, though I will admit to being disappointed that this plugin doesn't receive the benefit of being tested on the highly popular Python 3.4 and 3.5 platforms, and doubly so given that the reason is to avoid maintenance effort when "there is nothing in this repo that is likely to break across minor versions."

twolfson:
Cool, thank you for understanding

Issue #23, opened 2016-02-16
Package version 0.2.3 egg is broken


do3cc:
The egg does not contain the `VERSION` file.
the module tries to load the `VERSION` file during import time and fails.
I guess with this
https://github.com/zheller/flake8-quotes/blob/master/setup.py#L31

this line gets ignored:
https://github.com/zheller/flake8-quotes/blob/master/setup.py#L33

Ichimonji10:
See #21.

do3cc:
Pkg_resources really is the way to.

Release management really becomes a no brainer when following a few best practices documented and built upon in zest.releaser.

Von meinem iPhone gesendet

Am 16.02.2016 um 19:30 schrieb Ichimonji10 <notifications@github.com<mailto:notifications@github.com>>:


See #21<https://github.com/zheller/flake8-quotes/pull/21>.

-
Reply to this email directly or view it on GitHub<https://github.com/zheller/flake8-quotes/issues/23#issuecomment-184815564>.

________________________________________________________ The contents of this e-mail and any attachments are confidential to the intended recipient. They may not be disclosed to or used by or copied in any way by anyone other than the intended recipient. If this e-mail is received in error, please immediately notify the sender and delete the e-mail and attached documents. Please note that neither the sender nor the sender's company accept any responsibility for viruses and it is your responsibility to scan or otherwise check this e-mail and any attachments.

twolfson:
Sorry, I was under the impression @Ichimonji10 was taking care of this in #21 =/

Ichimonji10:
Feel free to re-open and merge #21.

twolfson:
Alright, it looks like #21 is off of the table. I will take a look at resolving this issue by the end of the weekend. In the meantime, please pin to a lower version:

```bash
pip install flake8-quotes<0.2.0,>=0.1.2
```

twolfson:
PR #24 is now open to resolve this issue

richrd:
Great to see this getting fixed! It's breaking my travis build :relaxed: :+1:

twolfson:
Decided to cowboy land #24 due to the lack of review from @zheller. It's been landed/released in `0.2.4`

PR #24, opened 2016-02-17
Relocated VERSION into flake8_quotes/__about__.py

twolfson:
As reported in #21 and #23, we are having issues with externally installed `flake8-quotes`. We tried a few solutions to get a `VERSION` working but they weren't sticking.

As a result, we went with the tested and true method used by `flake8-import-order`; using an `__about__.py` file.

https://github.com/public/flake8-import-order/blob/2ac7052a4e02b4a8a0125a106d87465a3b9fd688/setup.py

In this PR:

- Relocated `flake8_quotes` and `VERSION` into `flake8_quotes/__init__.py` and `flake8_quotes/__about__.py` respectively
- Updated `release.sh`
- Updated `setup.py` to use `__about__.py` as in `flake8-import-order`
- Removed unnecessary `setuptools` dependency from `setup.py`
- Replaced `py_modules` with `packages` (seems to be what `charlatan` and `flask` use)

**Notes:**

We were able to reproduce the issue locally via the following:

```bash
# Create temporary flake8-quotes install
mkvirtualenv flake8-quotes
git checkout master

# Build `sdist` file as we do in `release.sh`
python setup.py sdist

# Navigate to parent directory to assure no issues
cd ..

# Install distribution file
pip install flake8_quotes/dist/flake8-quotes-0.2.3.tar.gz

# Verify error is reproduced
flake8 flake8_quotes/

# Try/rinse/repeat with different edits and recompiles
```

/cc @zheller


do3cc:
Please consider [zest.releaser](http://zestreleaser.readthedocs.org/en/latest/) as an alternative.
This will make flake8-quotes much smaller, because you won't need a bunch of files any more

twolfson:
We have already explained our opposition to `pkg_resources`/`zest.releaser` in #21 and #23. That being said, I realized there's a simpler way to do this than needing an `__about__.py` (i.e. `splitlines` + filter + `assert`)

twolfson:
Rewrote the `about` logic as:

```
about = {}
flake8_quotes_lines = read(os.path.join(__dir__, 'flake8_quotes.py')).splitlines()
about_lines_beginnings = ['__version__ =']
about_lines = [
    line for line in flake8_quotes_lines if
    any([beginning for beginning in about_lines_beginnings if line.startswith(beginning)])
]
assert len(about_lines) == len(about_lines_beginnings)
exec('\n'.join(about_lines), about)
```

but it feels dirtier. @zheller Please review this PR when you get a chance

twolfson:
On a tangential issue I have discovered that `pypa` (people who run PyPI) use the `__about__.py` pattern which makes me much more confident in our decision

twolfson:
https://github.com/pypa/readme_renderer/tree/e78d4092026f0e67417109cf40329eadc5e88834/readme_renderer

eklitzke:
@twolfson Your approach has multiple errors:
* You're eagerly reading in the entire file with `read()` (also I think you mean `open().read()`, although I'm not up to snuff with the Python 3 stuff)
 * You're eagerly reading in the entire file with `.readlines()` which is inefficient in every possible way (memory, disk, cpu). You should use the normal iterator syntax, i.e. `for line in file:` instead of creating a (potentially very large) list object
 * You're using a list comprehension in the `any()` expression which makes it non-lazy.

```python
# BAD never do this; its slower, and allocates more memory since it creates
#  an intermediate list object
any([foo for foo in all_foos if some_condition(foo)])

# GOOD always do this; it's faster and allocates less memory because there is
#  no intermediate list object
any(foo for foo in all_foos if some_condition(foo))
```

Cheers,
Evan

do3cc:
@eklitzke You may want to reread this pull request again. two of your points don't occur at all. 
Your first point shows that you should read the python docs again about with statements: https://docs.python.org/2.7/reference/compound_stmts.html#with
Also, it makes a nicer read if you differentiate between errors and possible improvements.

twolfson:
@do3cc I believe @eklitzke was talking with respect to my inline example in https://github.com/zheller/flake8-quotes/pull/24#issuecomment-186518693

@eklitzke With respect to `read`, the example was using the `read` function defined in `setup.py` https://github.com/zheller/flake8-quotes/blob/0.2.3/setup.py#L5-L12. That being said, a `for ... in` to iterate over lines would definitely be more efficient. Nice catch.

With respect to `any`, I was considering using a generator expression but thought it might confuse people who don't know about them so I skipped them. Good catch though :+1:

twolfson:
Getting frustrated at this PR still being open. Going to cowboy land it since not having it is breaking more and more builds.

do3cc:
0.2.4 looks good :+1:

andrew-regan:
Same here :+1:

PR #26, opened 2016-04-29
Add support for defining multiline quote type

### LINKED ISSUE #25: Add support for docstrings
I'd like to be able to enforce docstring quotes and normal string quotes separately, so I can require double quotes for docstrings and single quotes everywhere else.

Example of code that would pass the check (assuming we enforce double quotes for docstrings and single quotes for everything else):

``` python
def hello():
    """
    Prints "Hello, world!"
    """
    print('Hello, world!')
```

And an example of some code that would not pass using the aforementioned rules:

``` python
def hello():
    '''
    Prints "Hello, world!"
    '''
    print('Hello, world!')
```

``` python
def hello():
    '''
    Prints "Hello, world!"
    '''
    print("Hello, world!")
```

``` python
def hello():
    """
    Prints "Hello, world!"
    """
    print("Hello, world!")
```

jackwilsdon:
This fixes #25.
- Improve naming and documentation of QUOTES and MULTILINE_QUOTES
  dictionaries inside `QuoteChecker`.
- Add new `--multiline-quotes` option to allow customisation of
  multiline quote type.
- Improve logic inside `QuoteChecker.get_quotes_errors` so that we can
  check single and multiline quotes separately.
- Add new tests for mixed strings and fix tests broken by the new
  multiline rules.

It seems there was a bug before that ignored multiline comments altogether which I have now fixed.

### DIFF
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -10,15 +10,13 @@ class QuoteChecker(object):
     version = __version__
 
     QUOTES = {
-        # Allow single quotes
-        # Fail on double quotes
+        # When user wants only single quotes
         '\'': {
             'allow': '\'',
             'fail': '"',
         },
 
-        # Allow double quotes
-        # Fail on single quotes
+        # When user wants only double quotes
         '"': {
             'allow': '"',
             'fail': '\'',
@@ -26,15 +24,13 @@ class QuoteChecker(object):
     }
 
     MULTILINE_QUOTES = {
-        # Allow single multiline quotes
-        # Fail on double multiline quotes
+        # When user wants only single multiline quotes
         '\'': {
             'allow': '\'\'\'',
             'fail': '"""'
         },
 
-        # Allow double multiline quotes
-        # Fail on single multiline quotes
+        # When user wants only double multiline quotes
         '"': {
             'allow': '"""',
             'fail': '\'\'\''
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -44,19 +44,27 @@ class QuoteChecker(object):
     def add_options(cls, parser):
         parser.add_option('--quotes', default='\'', action='store',
                           help='Quote to expect in all files (default: \')')
-        parser.add_option('--multiline-quotes', default='\'', action='store',
-                          help='Multiline quote to expect in all files (default: \')')
+        parser.add_option('--multiline-quotes', default='', action='store',
+                          help='Multiline quote to expect in all files (default: any)')
         parser.config_options.extend(['quotes', 'multiline_quotes'])
 
     @classmethod
     def parse_options(cls, options):
+        multiline_ignored = True
+
         if hasattr(options, 'multiline_quotes'):
             multiline_quotes = options.multiline_quotes
+
+            if len(multiline_quotes) == 0:
+                multiline_quotes = options.quotes
+            else:
+                multiline_ignored = False
         else:
             multiline_quotes = options.quotes
 
         cls.quotes = cls.INLINE_QUOTES[options.quotes]
         cls.multiline_quotes = cls.MULTILINE_QUOTES[multiline_quotes]
+        cls.multiline_ignored = multiline_ignored
 
     def get_file_contents(self):
         if self.filename in ('stdin', '-', None):
@@ -90,6 +98,10 @@ class QuoteChecker(object):
 
             if not token.string.startswith(self.quotes['bad']) and not token.string.startswith(self.multiline_quotes['bad']):
                 # ignore strings that do not start with our quotes (both single and multiline)
+                continue
+
+            if self.multiline_ignored and any([token.string.startswith(value) for value in self.multiline_quotes.values()]):
+                # ignore any type of multiline if multilines are being ignored
                 continue
 
             if self.quotes['good'] in token.string:
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -37,6 +37,8 @@ class QuoteChecker(object):
         }
     }
 
+    MULTILINE_QUOTES_STRINGS = ('\'\'\'', '"""')
+
     def __init__(self, tree, filename='(none)', builtins=None):
         self.filename = filename
 
@@ -50,21 +52,17 @@ class QuoteChecker(object):
 
     @classmethod
     def parse_options(cls, options):
-        multiline_ignored = True
-
         if hasattr(options, 'multiline_quotes'):
             multiline_quotes = options.multiline_quotes
-
-            if multiline_quotes is None:
-                multiline_quotes = options.inline_quotes
-            else:
-                multiline_ignored = False
         else:
             multiline_quotes = options.inline_quotes
 
         cls.inline_quotes = cls.INLINE_QUOTES[options.inline_quotes]
-        cls.multiline_quotes = cls.MULTILINE_QUOTES[multiline_quotes]
-        cls.multiline_ignored = multiline_ignored
+
+        if multiline_quotes is not None:
+            cls.multiline_quotes = cls.MULTILINE_QUOTES[multiline_quotes]
+        else:
+            cls.multiline_quotes = None
 
     def get_file_contents(self):
         if self.filename in ('stdin', '-', None):
@@ -96,13 +94,15 @@ class QuoteChecker(object):
                 # ignore non strings
                 continue
 
-            if not token.string.startswith(self.inline_quotes['bad']) and not token.string.startswith(self.multiline_quotes['bad']):
-                # ignore strings that do not start with our quotes (both single and multiline)
+            if token.string[0:3] in self.MULTILINE_QUOTES_STRINGS:
+                if self.multiline_quotes is None or token.string.startswith(self.multiline_quotes['bad']):
+                    continue
+
+            if not token.string.startswith(self.inline_quotes['bad']):
+                # ignore strings that do not start with our quote
                 continue
 
-            if self.multiline_ignored and any([token.string.startswith(value) for value in self.multiline_quotes.values()]):
-                # ignore any type of multiline if multilines are being ignored
-                continue
+
 
             if self.inline_quotes['good'] in token.string:
                 # ignore quotes wrapped in our quotes (e.g. `'` in `"it's"`)
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -9,7 +9,7 @@ class QuoteChecker(object):
     name = __name__
     version = __version__
 
-    QUOTES = {
+    INLINE_QUOTES = {
         # When user wants only single quotes
         '\'': {
             'good': '\'',
@@ -55,7 +55,7 @@ class QuoteChecker(object):
         else:
             multiline_quotes = options.quotes
 
-        cls.quotes = cls.QUOTES[options.quotes]
+        cls.quotes = cls.INLINE_QUOTES[options.quotes]
         cls.multiline_quotes = cls.MULTILINE_QUOTES[multiline_quotes]
 
     def get_file_contents(self):
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -12,28 +12,28 @@ class QuoteChecker(object):
     QUOTES = {
         # When user wants only single quotes
         '\'': {
-            'allow': '\'',
-            'fail': '"',
+            'good': '\'',
+            'bad': '"',
         },
 
         # When user wants only double quotes
         '"': {
-            'allow': '"',
-            'fail': '\'',
+            'good': '"',
+            'bad': '\'',
         },
     }
 
     MULTILINE_QUOTES = {
         # When user wants only single multiline quotes
         '\'': {
-            'allow': '\'\'\'',
-            'fail': '"""'
+            'good': '\'\'\'',
+            'bad': '"""'
         },
 
         # When user wants only double multiline quotes
         '"': {
-            'allow': '"""',
-            'fail': '\'\'\''
+            'good': '"""',
+            'bad': '\'\'\''
         }
     }
 
@@ -88,15 +88,15 @@ class QuoteChecker(object):
                 # ignore non strings
                 continue
 
-            if not token.string.startswith(self.quotes['fail']) and not token.string.startswith(self.multiline_quotes['fail']):
+            if not token.string.startswith(self.quotes['bad']) and not token.string.startswith(self.multiline_quotes['bad']):
                 # ignore strings that do not start with our quotes (both single and multiline)
                 continue
 
-            if self.quotes['allow'] in token.string:
+            if self.quotes['good'] in token.string:
                 # ignore quotes wrapped in our quotes (e.g. `'` in `"it's"`)
                 continue
 
-            if self.multiline_quotes['allow'] in token.string:
+            if self.multiline_quotes['good'] in token.string:
                 # ignore multiline quotes wrapped in our quotes
                 continue
 
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -44,7 +44,7 @@ class QuoteChecker(object):
     def add_options(cls, parser):
         parser.add_option('--quotes', default='\'', action='store',
                           help='Quote to expect in all files (default: \')')
-        parser.add_option('--multiline-quotes', default='', action='store',
+        parser.add_option('--multiline-quotes', action='store',
                           help='Multiline quote to expect in all files (default: any)')
         parser.config_options.extend(['quotes', 'multiline_quotes'])
 
@@ -55,7 +55,7 @@ class QuoteChecker(object):
         if hasattr(options, 'multiline_quotes'):
             multiline_quotes = options.multiline_quotes
 
-            if len(multiline_quotes) == 0:
+            if multiline_quotes is None:
                 multiline_quotes = options.quotes
             else:
                 multiline_ignored = False
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -9,7 +9,7 @@ class QuoteChecker(object):
     name = __name__
     version = __version__
 
-    INLINE_QUOTES = {
+    QUOTES = {
         # When user wants only single quotes
         '\'': {
             'good': '\'',
@@ -44,7 +44,7 @@ class QuoteChecker(object):
 
     @classmethod
     def add_options(cls, parser):
-        parser.add_option('--quotes', default='\'', action='store', dest='inline_quotes',
+        parser.add_option('--quotes', default='\'', action='store',
                           help='Quote to expect in all files (default: \')')
         parser.add_option('--multiline-quotes', action='store',
                           help='Multiline quote to expect in all files (disabled by default)')
@@ -55,9 +55,9 @@ class QuoteChecker(object):
         if hasattr(options, 'multiline_quotes'):
             multiline_quotes = options.multiline_quotes
         else:
-            multiline_quotes = options.inline_quotes
+            multiline_quotes = options.quotes
 
-        cls.inline_quotes = cls.INLINE_QUOTES[options.inline_quotes]
+        cls.quotes = cls.QUOTES[options.quotes]
         cls.multiline_quotes = cls.MULTILINE_QUOTES[multiline_quotes] if multiline_quotes is not None else None
 
     def get_file_contents(self):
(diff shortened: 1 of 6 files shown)


flake8_quotes/__init__.py
```diff
@@ -10,33 +10,57 @@ class QuoteChecker(object):
...
 
     @classmethod
     def parse_options(cls, options):
+        if hasattr(options, 'multiline_quotes'):
+            multiline_quotes = options.multiline_quotes
+        else:
+            multiline_quotes = options.quotes
+
         cls.quotes = cls.QUOTES[options.quotes]
+        cls.multiline_quotes = cls.MULTILINE_QUOTES[multiline_quotes]
```
twolfson: This is going to break existing behavior (e.g. people might be using single quotes for inline strings and double quotes for multiline strings). Maybe we should allow for a `None` option?

flake8_quotes/__init__.py
```diff
@@ -10,33 +10,57 @@ class QuoteChecker(object):
     version = __version__
 
     QUOTES = {
-        # When user wants only single quotes
+        # Allow single quotes
+        # Fail on double quotes
```
twolfson: This comment is redundant to the content, the old comment stated the scenario when the values were used =/

flake8_quotes/__init__.py
jackwilsdon: Is this related to the bug I noted in the PR description? If so I was considering fixing it as modifying the tests felt a bit dirty and I thought it might break some stuff for existing users. I'll add that option now.

flake8_quotes/__init__.py
```diff
@@ -10,33 +10,57 @@ class QuoteChecker(object):
...
         '\'': {
-            'good_single': '\'',
-            'good_multiline': '\'\'\'',
-            'bad_single': '"',
-            'bad_multiline': '"""',
+            'allow': '\'',
+            'fail': '"',
```
twolfson: Can we move back to `good/bad`? It will cause less PR noise and they were adjectives describing the strings (which makes sense in conditionals), not verbs describing actions to perform against strings (makes less sense in conditionals unless they are methods).

twolfson:
At a glance, everything seems reasonable. I left some comments about breaking backwards compatibility and unnecessary PR noise.

flake8_quotes/__init__.py
jackwilsdon: Sure, I only changed it because I found it a bit confusing at first 😄.

flake8_quotes/__init__.py
twolfson: The current behavior is no linting of multiline strings. We cannot assume single or double quotes to be the default without breaking backwards compatibility. As a result, we should decide to add support for ignoring them (preferred) or make this as part of a major release (less preferred).

flake8_quotes/__init__.py
```diff
@@ -12,31 +12,51 @@ class QuoteChecker(object):
     QUOTES = {
```
twolfson: Maybe we should rename this to `INLINE_QUOTES` as well as the other references to avoid ambiguity?

jackwilsdon:
I'm not too pleased with the implementation of how multiline quotes are ignored but I'm not too sure on the best way to improve it.

flake8_quotes/__init__.py
```diff
@@ -44,19 +44,27 @@ def __init__(self, tree, filename='(none)', builtins=None):
     def add_options(cls, parser):
         parser.add_option('--quotes', default='\'', action='store',
                           help='Quote to expect in all files (default: \')')
-        parser.add_option('--multiline-quotes', default='\'', action='store',
-                          help='Multiline quote to expect in all files (default: \')')
+        parser.add_option('--multiline-quotes', default='', action='store',
```
twolfson: We should be able to remove `default` and assume `None` when nothing is present

flake8_quotes/__init__.py
```diff
@@ -44,19 +44,27 @@ def __init__(self, tree, filename='(none)', builtins=None):
...
+                          help='Multiline quote to expect in all files (default: any)')
```
twolfson: As a user, "default: any" would make me assume `--multiline-quotes any` would work =/ Maybe we should say `...in all files (e.g. """, ''')`

flake8_quotes/__init__.py
```diff
@@ -44,19 +44,27 @@ def __init__(self, tree, filename='(none)', builtins=None):
...
     def parse_options(cls, options):
+        multiline_ignored = True
+
         if hasattr(options, 'multiline_quotes'):
             multiline_quotes = options.multiline_quotes
+
+            if len(multiline_quotes) == 0:
+                multiline_quotes = options.quotes
+            else:
+                multiline_ignored = False
```
twolfson: Instead of breaking up quotes/ignored into 2 separate variables, how about we allow `multiline_quotes` to be `None`? Then we can do `if self.multiline_quotes and self.multiline_quotes['good'] ...` in our logic

flake8_quotes/__init__.py
```diff
@@ -44,19 +44,27 @@ def __init__(self, tree, filename='(none)', builtins=None):
...
         else:
             multiline_quotes = options.quotes
 
-        cls.quotes = cls.QUOTES[options.quotes]
+        cls.quotes = cls.INLINE_QUOTES[options.quotes]
```
twolfson: For consistency, this should be `cls.inline_quotes` as well as our other references to anything like quotes (e.g. `tests`, CLI option). However, maybe that will be too noisy for this PR. Thoughts?

flake8_quotes/__init__.py
jackwilsdon: I did it this way as we need a reference to all of the possible `multiline_quotes` so that we can check if the current token is a multiline quote and we can ignore it if needed.

flake8_quotes/__init__.py
jackwilsdon: It does seem a bit outside the scope of this PR but we may as well do it for consistency, I'll update it now.

flake8_quotes/__init__.py
```diff
@@ -92,6 +100,10 @@ def get_quotes_errors(self, file_contents):
                 # ignore strings that do not start with our quotes (both single and multiline)
                 continue
 
+            if self.multiline_ignored and any([token.string.startswith(value) for value in self.multiline_quotes.values()]):
```
twolfson: Instead of an any check, maybe we can simplify the logic to only use 'good'?

```python
token.string[0:3] == self.multiline_quotes['good']
```

flake8_quotes/__init__.py
twolfson: I believe `argparse` supports aliases so we should be fine

flake8_quotes/__init__.py
jackwilsdon: Doing that causes any tokens with "bad" multiline quotes (even ignored ones) to fail.

flake8_quotes/__init__.py
twolfson: Ah, I think the logic was a bit fucky there. We shouldn't care about the user's option if they want us to ignore it. Let's try the following:

```
# Define a new constant named `MULTILINE_QUOTE_STRS
# This should be after MULTILINE_QUOTES
MULTILINE_QUOTE_STRS = ('"""', '\'\'\'')

# Replace the logic with something like
# If we have a multiline string
if token.string[0:3] in self.MULTILINE_QUOTE_STRS:
    # If we are ignoring multiline strings, then skip it
    if not self.multiline_quotes:
        continue
    # Otherwise, validate it
    elif token.string.startswith(self.multiline_quotes['good']):
      continue
```

flake8_quotes/__init__.py
twolfson: See updated logic suggestion here:

https://github.com/zheller/flake8-quotes/pull/26/files/22ed02b11ad1883a3e404269a88d3669bb1875ac..3fda3d44e7fcf43a5173383a24310ac2c8d56915#r61526232

flake8_quotes/__init__.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, tree, filename='(none)', builtins=None):
 
     @classmethod
     def add_options(cls, parser):
-        parser.add_option('--quotes', default='\'', action='store',
+        parser.add_option('--quotes', default='\'', action='store', dest='inline_quotes',
```
twolfson: We should prob name it `--inline-quotes` too =/

twolfson: So maybe make 2 CLI options (`--quotes` and `--inline-quotes`) with `--quotes` as deprecated. This is prob getting too big for this PR. Maybe it should be a precursor PR?

jackwilsdon: Ah, that's what you meant.

flake8_quotes/__init__.py
```diff
@@ -68,16 +94,22 @@ def get_quotes_errors(self, file_contents):
                 # ignore non strings
                 continue
 
-            if not token.string.startswith(self.quotes['bad_single']):
-                # ignore strings that do not start with our quotes
+            if token.string[0:3] in self.MULTILINE_QUOTES_STRINGS:
+                if self.multiline_quotes is None or token.string.startswith(self.multiline_quotes['bad']):
```
twolfson: Please add a comment explaining what we are skipping

flake8_quotes/__init__.py
```diff
@@ -68,16 +94,22 @@ def get_quotes_errors(self, file_contents):
...
+                    continue
+
+            if not token.string.startswith(self.inline_quotes['bad']):
+                # ignore strings that do not start with our quote
                 continue
 
-            if token.string.startswith(self.quotes['bad_multiline']):
-                # ignore multiline strings
+
+
```
twolfson: The linter will prob complain about this but there's too much whitespace here =/

flake8_quotes/__init__.py
```diff
@@ -9,34 +9,60 @@ class QuoteChecker(object):
...
+            multiline_quotes = options.multiline_quotes
+        else:
+            multiline_quotes = options.inline_quotes
+
+        cls.inline_quotes = cls.INLINE_QUOTES[options.inline_quotes]
+
+        if multiline_quotes is not None:
+            cls.multiline_quotes = cls.MULTILINE_QUOTES[multiline_quotes]
+        else:
+            cls.multiline_quotes = None
```
twolfson: I think this logic is currently too spread out for multiline quotes. It should look something like:

```python
def parse_options(cls, options):
  if hasattr(options, 'multiline_quotes'):
    cls.multiline_quotes = cls.MULTILINE_QUOTES[options.multiline_quotes]
  else:
    cls.multiline_quotes = None

  cls.inline_quotes = cls.INLINE_QUOTES[options.inline_quotes]
```

jackwilsdon:
Should I just use `git revert` to revert 3191735d918e04092f2ed87d033f0eb7506aebef, c03a098defe958d78f4409f6b3d34c42bcb9af62 and 5892a57e72ea52c3dee94299abe58e62ebcfa7f1? (the three commits that renamed `quotes` to `inline_quotes`).

twolfson:
Yea, you can use `git revert` but there's probably some intermingling of content. I suggest using `git checkout -p` which allows you to check out patches from a previous commit. For example:

```bash
git checkout -p 22ed02b
```

flake8_quotes/__init__.py
```diff
@@ -68,16 +90,20 @@ def get_quotes_errors(self, file_contents):
...
-                # ignore multiline strings
+            if self.quotes['good'] in token.string:
+                # ignore quotes wrapped in our quotes (e.g. `'` in `"it's"`)
                 continue
 
-            if self.quotes['good_single'] in token.string:
-                # ignore alternate quotes wrapped in our quotes (e.g. `'` in `"it's"`)
+            if self.multiline_quotes['good'] in token.string:
+                # ignore multiline quotes wrapped in our quotes
                 continue
```
twolfson: We prob don't need to exempt strings containing multiline quotes (e.g. `"""` in `'its"""'`)

test/test_checks.py
```diff
@@ -34,6 +34,18 @@ def test_noqa_doubles(self):
         self.assertEqual(list(checker.run()), [])
 
 
+class DoublesMixedTestChecks(TestCase):
```
twolfson: It's ambiguous what mixed means. Maybe we should rename to `DoublesInlineSinglesMultilineTestChecks` and similar for other suites?

twolfson: However, we might be better off keeping inline tests separate from multiline tests altogether (e.g. `SinglesMultilineTestChecks`)

jackwilsdon: That sounds like another PR coming on to me 😆.

twolfson: Nah, it should be fine for here. #27 should be landed soon and then all multiline alterations belong here =P

xZise:
Afaik should this use a different error code.

flake8_quotes/__init__.py
```diff
@@ -12,31 +12,53 @@ class QuoteChecker(object):
...
 
     @classmethod
     def parse_options(cls, options):
+        if hasattr(options, 'multiline_quotes'):
+            multiline_quotes = options.multiline_quotes
+        else:
+            multiline_quotes = None
+
         cls.quotes = cls.QUOTES[options.quotes]
+        cls.multiline_quotes = cls.MULTILINE_QUOTES[multiline_quotes] if multiline_quotes is not None else None
```
xZise: If you use a different error code, you can just use the quotes selected and then you don't have to worry about it being None blow (just here).

flake8_quotes/__init__.py
```diff
@@ -12,31 +12,53 @@ class QuoteChecker(object):
...
         parser.add_option('--quotes', default='\'', action='store',
                           help='Quote to expect in all files (default: \')')
-        parser.config_options.append('quotes')
+        parser.add_option('--multiline-quotes', action='store',
+                          help='Multiline quote to expect in all files (disabled by default)')
+        parser.config_options.extend(['quotes', 'multiline_quotes'])
 
     @classmethod
     def parse_options(cls, options):
+        if hasattr(options, 'multiline_quotes'):
```
xZise: Is there any reason this is `False` apart from the tests? In that case the tests should provide it (see my patch #33), as the production code shouldn't be modified just because of tests and instead the tests should be modified (if possible and reasonable).

jackwilsdon: Is there any point in me completing this PR, as #33 seems to greatly supersede this one.

twolfson: I would be more comfortable with landing this PR (#26) over #33 since it has been discussed for a while and doesn't cause too much restructuring of the tool for an enhancement. Support for things like new error codes would be best addressed by a separate issue.

twolfson:
I'm pretty sure this PR is more/less done except for renaming test suites. Can you take care of that?

jackwilsdon:
Would you advise a rebase or a merge from master? The only issue I can see with a rebase is the possibility of losing the diff discussions from earlier in the thread.

jackwilsdon:
Another thing, what do you think of renaming most of the tests to be a bit more explicit? (i.e. explaining the expected outcome in the name)

Here's an example of what I mean:

```diff
-class DoublesTestChecks(TestCase):
+class InlineDoublesTestChecks(TestCase):
     def setUp(self):
         class DoublesOptions():
             quotes = '\''
         QuoteChecker.parse_options(DoublesOptions)
 
-    def test_multiline_string(self):
+    def test_multiline_strings_are_not_linted(self):
         doubles_checker = QuoteChecker(None, filename=get_absolute_path('data/doubles_multiline_string.py'))
         self.assertEqual(list(doubles_checker.get_quotes_errors(doubles_checker.get_file_contents())), [])
 
-    def test_wrapped(self):
+    def test_wrapped_are_not_linted(self):
         doubles_checker = QuoteChecker(None, filename=get_absolute_path('data/doubles_wrapped.py'))
         self.assertEqual(list(doubles_checker.get_quotes_errors(doubles_checker.get_file_contents())), [])
 
-    def test_doubles(self):
+    def test_doubles_are_linted(self):
         doubles_checker = QuoteChecker(None, filename=get_absolute_path('data/doubles.py'))
         self.assertEqual(list(doubles_checker.get_quotes_errors(doubles_checker.get_file_contents())), [
             {'col': 24, 'line': 1, 'message': 'Q000 Remove bad quotes.'}
         ])
 
-    def test_noqa_doubles(self):
+    def test_noqa_doubles_are_not_linted(self):
         checker = QuoteChecker(None, get_absolute_path('data/doubles_noqa.py'))
         self.assertEqual(list(checker.run()), [])
```

flake8_quotes/__init__.py
```diff
@@ -12,31 +12,53 @@ class QuoteChecker(object):
...
-            'good_single': '"',
-            'good_multiline': '"""',
-            'bad_single': '\'',
-            'bad_multiline': '\'\'\'',
+            'good': '"',
+            'bad': '\'',
         },
     }
 
+    MULTILINE_QUOTES = {
```
xZise: Maybe this could be already implemented similar to how I did it in #33 and use a list. I don't see the advantage of using the dicts.

xZise:
I personally prefer rebasing/squashing changes, as I like my repos to be clean (unless it is some official branch like master and public). Now this is not my repo so I can't speak for @twolfson but Github should still be able to show comments on older commits. For example #32 has a removed commit and you can still see the comments (you have to click on “Show comments” not “outdated diff”).

And I think renaming the methods could be done in a different patch as they are independent. But renaming the classes could be done here.

twolfson:
Sorry for taking so long to get back to this, it's been an unexpectedly hectic weekend. I will reply to comments on this and other threads by the end of the week.

flake8_quotes/__init__.py
jackwilsdon: I just used dicts as that's what was being used for `QUOTES`, I find it makes it easier to read and there's no real advantage of using a list either.

jackwilsdon:
@xZise I was talking about resolving the existing merge conflict more than squashing my commits, although I'm sure I will end up squashing them anyway.

twolfson:
Sorry, I misjudged my timing; some things popped up which killed more of my time than I anticipated. I am bumping the ETA to the end of next week =/

jackwilsdon:
That's fine, no worries @twolfson!

xZise:
As I personally prefer rebasing over merging, I'd rebase these changes onto master and squash them.

flake8_quotes/__init__.py
xZise: Main reason I dislike the current situation (also `QUOTES`) is that it contains redundant data. X's good is Y's bad and Y's good is X's bad. Also if there are really more quote styles, a list is handier as you could just remove the accepted style and have all bad ones. Especially with this patch using dicts doesn't really convey much more information apart from having a mapping from the setting value (e.g. `'`) to the actual value (`'''`). Also you do use a list already with `MULTILINE_QUOTES_STRINGS`.

But as it seems my patch ( #33 ) should be based on this one, I could implement it later.

jackwilsdon: We can leave it to @twolfson to decide on this, I just attempted to mirror the existing implementation of `QUOTES`.

twolfson: I'm positive we discussed this at length on this PR previously; we want to keep them as dicts to reduce the amount of noise in the PR. If we want to change implementation of existing components, then that should be discussed/done in another issue/PR.

flake8_quotes/__init__.py
twolfson: We allow multiline quotes to not be defined so we can maintain backwards compatibility and release this PR as a minor revision.

flake8_quotes/__init__.py
twolfson: This is an arguable point but I feel that introducing a new error code/behavior would be backwards incompatible behavior and should be done as a major release which is not the intent of this PR

twolfson:
@jackwilsdon With respect to https://github.com/zheller/flake8-quotes/pull/26#issuecomment-231512686, I think we should pass on renaming tests for now. It will add unnecessary noise to the PR which gives us more to review and is trivial to move to another PR.

twolfson:
@jackwilsdon I'm not sure of the local status for your git repository but you should be able to do:

```bash
# Rebase/squash branch onto original `master`
git rebase -i 778a4db0e9959f56f6f416ae3039c1b78f61117e

# Rebase single commit onto latest master
git rebase origin/master
```

twolfson:
It's been a while since this PR has been touched and I've lost most context in my memory. Going to close this PR due to inactivity. If someone wants to pick this back up, please let us know and we'll re-open or open another PR

PR #27, opened 2016-04-29
Rename quotes option

jackwilsdon:
Rename `--quotes` to `--inline-quotes` as suggested in #26.

I'm not too sure if the best way to mark it as deprecated is to just put "(deprecated)" after the command, what do you think @twolfson?

### DIFF
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -9,7 +9,7 @@ class QuoteChecker(object):
     name = __name__
     version = __version__
 
-    QUOTES = {
+    INLINE_QUOTES = {
         # When user wants only single quotes
         '\'': {
             'good_single': '\'',
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -30,13 +30,13 @@ class QuoteChecker(object):
 
     @classmethod
     def add_options(cls, parser):
-        parser.add_option('--quotes', default='\'', action='store',
+        parser.add_option('--quotes', default='\'', action='store', dest='inline_quotes',
                           help='Quote to expect in all files (default: \')')
         parser.config_options.append('quotes')
 
     @classmethod
     def parse_options(cls, options):
-        cls.quotes = cls.QUOTES[options.quotes]
+        cls.inline_quotes = cls.INLINE_QUOTES[options.inline_quotes]
 
     def get_file_contents(self):
         if self.filename in ('stdin', '-', None):
@@ -68,15 +68,15 @@ class QuoteChecker(object):
                 # ignore non strings
                 continue
 
-            if not token.string.startswith(self.quotes['bad_single']):
+            if not token.string.startswith(self.inline_quotes['bad_single']):
                 # ignore strings that do not start with our quotes
                 continue
 
-            if token.string.startswith(self.quotes['bad_multiline']):
+            if token.string.startswith(self.inline_quotes['bad_multiline']):
                 # ignore multiline strings
                 continue
 
-            if self.quotes['good_single'] in token.string:
+            if self.inline_quotes['good_single'] in token.string:
                 # ignore alternate quotes wrapped in our quotes (e.g. `'` in `"it's"`)
                 continue
 
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -30,8 +30,8 @@ class QuoteChecker(object):
 
     @classmethod
     def add_options(cls, parser):
-        parser.add_option('--quotes', default='\'', action='store', dest='inline_quotes',
-                          help='Quote to expect in all files (default: \') (deprecated)')
+        parser.add_option('--quotes', action='store', dest='inline_quotes',
+                          help='Deprecated alias for `--inline-quotes`')
         parser.add_option('--inline-quotes', default='\'', action='store',
                           help='Quote to expect in all files (default: \')')
         parser.config_options.append('quotes')
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -30,7 +30,7 @@ class QuoteChecker(object):
 
     @classmethod
     def add_options(cls, parser):
-        parser.add_option('--quotes', action='store', dest='inline_quotes',
+        parser.add_option('--quotes', action='store',
                           help='Deprecated alias for `--inline-quotes`')
         parser.add_option('--inline-quotes', default='\'', action='store',
                           help='Quote to expect in all files (default: \')')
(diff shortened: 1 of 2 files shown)


twolfson:
Woot, thanks for opening this as a separate PR =)

twolfson:
With respect to "deprecated", it's prob best to update the description altogether (i.e. mention what it was deprecated in favor of). For example:

```
optional arguments:
  -h, --help           show this help message and exit
  --quotes             Deprecated alias for `--inline-quotes`
  --inline-quotes      Quote to expect in all files (default: ')
```

flake8_quotes/__init__.py
```diff
@@ -30,13 +30,15 @@ def __init__(self, tree, filename='(none)', builtins=None):
 
     @classmethod
     def add_options(cls, parser):
-        parser.add_option('--quotes', default='\'', action='store',
+        parser.add_option('--quotes', default='\'', action='store', dest='inline_quotes',
+                          help='Quote to expect in all files (default: \') (deprecated)')
+        parser.add_option('--inline-quotes', default='\'', action='store',
                           help='Quote to expect in all files (default: \')')
         parser.config_options.append('quotes')
```
twolfson: We need to add `inline_quotes` for our config options here

flake8_quotes/__init__.py
```diff
@@ -30,13 +30,15 @@ def __init__(self, tree, filename='(none)', builtins=None):
 
     @classmethod
     def add_options(cls, parser):
-        parser.add_option('--quotes', default='\'', action='store',
+        parser.add_option('--quotes', default='\'', action='store', dest='inline_quotes',
```
twolfson: We can prob skip the `dest` here since we will need override logic for a `setup.cfg`

flake8_quotes/__init__.py
```diff
@@ -30,13 +30,15 @@ def __init__(self, tree, filename='(none)', builtins=None):
...
+                          help='Quote to expect in all files (default: \') (deprecated)')
+        parser.add_option('--inline-quotes', default='\'', action='store',
                           help='Quote to expect in all files (default: \')')
         parser.config_options.append('quotes')
 
     @classmethod
     def parse_options(cls, options):
-        cls.quotes = cls.QUOTES[options.quotes]
+        cls.inline_quotes = cls.INLINE_QUOTES[options.inline_quotes]
```
twolfson: We need to add a fallback to use `options.quotes` in case `options.inline_quotes` isn't defined

flake8_quotes/__init__.py
twolfson: We should also remove the default from here =/

flake8_quotes/__init__.py
jackwilsdon: I believe any hyphens are automatically converted to underscores, so the destination is not required (it ends up being `inline_quotes` anyway).

twolfson: `config_options` isn't the same as the CLI parser. `flake8` supports input via both CLI flags and configuration files (e.g. `setup.cfg`). `add_option` is for the former, `config_options` is for the latter.

flake8_quotes/__init__.py
jackwilsdon: Added in 9877c12585759c4d0b2269a75c8797f0ff83f76c.

flake8_quotes/__init__.py
jackwilsdon: I think the combination of 9877c12585759c4d0b2269a75c8797f0ff83f76c and d6dad0be845ecb0950109399474e2c629d2fa532 means both `quotes` and `inline-quotes` are valid options for flake8 and should work as expected?

flake8_quotes/__init__.py
jackwilsdon: Ah, I misread your comment. Fixed in d6dad0be845ecb0950109399474e2c629d2fa532.

flake8_quotes/__init__.py
twolfson: Yep, but if we have `--inline-quotes` as the primary option and someone was using the default already, then there should be no issue with there being no default set on `--quotes`.

jackwilsdon: I was referring to your first note, I removed the default in 2eb51c57b75966ce124a0b9283a39abc184c1a12 😁.

twolfson: Ah k. With respect to `dest`, it's a similar sentiment. We already have override logic to handle the configuration file case (i.e. `def parse_options`). Having an extra kwarg here (i.e. `dest`) is 1 more thing to go wrong/handle so we should consolidate them via `def parse_options`

flake8_quotes/__init__.py
```diff
@@ -30,13 +30,18 @@ def __init__(self, tree, filename='(none)', builtins=None):
 
     @classmethod
     def add_options(cls, parser):
-        parser.add_option('--quotes', default='\'', action='store',
+        parser.add_option('--quotes', action='store', dest='inline_quotes',
```
twolfson: Still waiting on the removal of `dest` from here =/

jackwilsdon: Ah, I completely forgot. Removed in b0129c4fb9af4bf8e25fde3226ca57b5a2411e3b.

flake8_quotes/__init__.py
```diff
@@ -30,13 +30,18 @@ def __init__(self, tree, filename='(none)', builtins=None):
...
+                          help='Deprecated alias for `--inline-quotes`')
+        parser.add_option('--inline-quotes', default='\'', action='store',
                           help='Quote to expect in all files (default: \')')
-        parser.config_options.append('quotes')
+        parser.config_options.extend(['quotes', 'inline-quotes'])
 
     @classmethod
     def parse_options(cls, options):
-        cls.quotes = cls.QUOTES[options.quotes]
+        if not hasattr(options, 'inline_quotes') and hasattr(options, 'quotes'):
```
twolfson: I was trying out the PR locally, the logic here is flawed as this is never true (i.e. `options.inline_quotes` is always defined since we have a default). Please update the logic to `hasattr(options, 'quotes')` only

jackwilsdon: Ah yes, I completely forgot about `inline_quotes` having a default! Thanks 👍

twolfson: Further events: `hasattr(options, 'quotes')` seems to always be `True` so we should prob use `if options.quotes` to check for falsy

jackwilsdon: Should be fixed in fe08c2cd2d8921cb111f0ced31b0804a382905cb. It turns out `hasattr(options, 'quotes')` isn't always `True` because our tests call the method directly 😆. Should be fixed now though.

twolfson:
Alright, looks good to me. Going to land this via a squash and release

twolfson:
This has been merged/released in `0.3.0` :tada:

jackwilsdon:
🎉 🎉 🎉 😁

Issue #30, opened 2016-06-29
using stdin no longer works


jdavidheiser:
Running flake8 with the following syntax

`cat file.py | flake8 -`

does not work with flake8-quotes, but does work with other packages.  This is most likely caused by the fact that recent versions of flake8 are now importing pycodestyle and not pep8.

xZise:
I get the following error:
```
$ cat flake8_quotes/__init__.py | flake8 -
Traceback (most recent call last):
  File "/home/xzise/.pyenv/versions/3.5.2/bin/flake8", line 11, in <module>
    sys.exit(main())
  File "/home/xzise/.pyenv/versions/3.5.2/lib/python3.5/site-packages/flake8/main.py", line 36, in main
    report = flake8_style.check_files()
  File "/home/xzise/.pyenv/versions/3.5.2/lib/python3.5/site-packages/flake8/engine.py", line 181, in check_files
    return self._retry_serial(self._styleguide.check_files, paths=paths)
  File "/home/xzise/.pyenv/versions/3.5.2/lib/python3.5/site-packages/flake8/engine.py", line 172, in _retry_serial
    return func(*args, **kwargs)
  File "/home/xzise/.pyenv/versions/3.5.2/lib/python3.5/site-packages/pycodestyle.py", line 1877, in check_files
    runner(path)
  File "/home/xzise/.pyenv/versions/3.5.2/lib/python3.5/site-packages/flake8/engine.py", line 126, in input_file
    return fchecker.check_all(expected=expected, line_offset=line_offset)
  File "/home/xzise/.pyenv/versions/3.5.2/lib/python3.5/site-packages/pycodestyle.py", line 1608, in check_all
    self.check_ast()
  File "/home/xzise/.pyenv/versions/3.5.2/lib/python3.5/site-packages/pycodestyle.py", line 1555, in check_ast
    for lineno, offset, text, check in checker.run():
  File "/home/xzise/Programms/flake8-quotes/flake8_quotes/__init__.py", line 57, in run
    file_contents = self.get_file_contents()
  File "/home/xzise/Programms/flake8-quotes/flake8_quotes/__init__.py", line 52, in get_file_contents
    return pep8.stdin_get_value().splitlines(True)
  File "/home/xzise/.pyenv/versions/3.5.2/lib/python3.5/site-packages/pep8-1.7.0-py3.5.egg/pep8.py", line 1198, in stdin_get_value
[...]

PR #31, opened 2016-06-29
Replace deprecated pep8 package with pycodestyle

jdavidheiser:
fix for issue #30, where the pep8 package doesn't work with parsing python files on stdin anymore.

### DIFF
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -1,7 +1,7 @@
 import tokenize
 import warnings
 
-import pep8
+import pycodestyle
 
 from flake8_quotes.__about__ import __version__
 
@@ -49,9 +49,9 @@ class QuoteChecker(object):
 
     def get_file_contents(self):
         if self.filename in ('stdin', '-', None):
-            return pep8.stdin_get_value().splitlines(True)
+            return pycodestyle.stdin_get_value().splitlines(True)
         else:
-            return pep8.readlines(self.filename)
+            return pycodestyle.readlines(self.filename)
 
     def run(self):
         file_contents = self.get_file_contents()
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(
     author_email='zheller@gmail.com',
     version=about['__version__'],
     install_requires=[
-        'pep8',
+        'pycodestyle',
     ],
     url='http://github.com/zheller/flake8-quotes/',
     long_description=LONG_DESCRIPTION,


twolfson:
It looks like `flake8` wants us to request the logical lines as our plugin. Can you explore this venue instead?

http://flake8.pycqa.org/en/latest/plugin-development/plugin-parameters.html

jdavidheiser:
Most other flake8 plugins aren't using the tree yet, and to be honest I don't have the time to restructure the logic of the project.  This simple fix works for the time being, and is at least on par with how other plugins are still doing it.

twolfson:
Could you link us to these other plugins?

xZise:
Apart from that it should either import `pep8` or `pycodestyle`, because if you installed `flake8` before the name change, this would break it.

xZise:
Okay I did a few tests and xZise/flake8-quotes@cc444c5 is using the import from `flake8` itself, which is `pycodestyle` on newer versions (like 2.6.2). This is fixing the issue I reported in #30 and it correctly checks the input. But when I rebase it on top #34 and use it with 3.0.0b2 it doesn't work because the read string from `stdin` is empty.

So I'm not sure how we can go forward here. Either it doesn't support 3.x using `stdin`, or there is some kind of backport (`FileProcessor` I guess?) which makes the basic `stdin` handling available to Flake8 2.x.

twolfson:
I don't understand how #34 would break it if it works in `3.x`. Fwiw, if you read directly from `stdin` (e.g. `sys.stdin.read()`) it will be empty as `flake8` has already consumed all the bytes into memory.

xZise:
Okay sorry for the spam, but I think I figured it out and will provide one (or maybe two) pull requests.

On Flake8 3.x we could actually use a custom constructor which directly gets us the lines, so the plugin doesn't need the pep8 magic. As it should support 2.x the pep8 magic would still be there but only run on 3.x. This is actually using `FileProcessor` (any parameter in the constructor is filled with the data from the attribute with the same name in `FileProcessor`) so in theory we could also use the tokens. I'll take a look into it.

The implementation I got was to have one base class and two subclasses depending on the Flake8 version. One just gets the lines (or tokens, and because of [Flake8 bug 159](https://gitlab.com/pycqa/flake8/issues/159) also `tree` but that isn't used). And the other subclass does the same as the current one as it reads from `stdin` or the file system and writes it into the lines (just as the 3.x version).

Additionally I figured out why on Flake8 the errors didn't show up (for me at least) as the errors aren't selected by default (and thus ignored). I'm not sure if this intentional or a bug and maybe I'll report a bug about that.

twolfson:
I don't understand why we would want to work around the simple interface that `pep8` provides. Can you elaborate?

xZise:
Didn't saw the new comment. #34 doesn't directly breaks it, but that enables this plugin to be used in 3.x. And there Flake8 seems to do it differently and consumes `stdin` before `pycodestyle` got to it, and when the plugin requests the `stdin` from `pycodestyle` it got nothing. It is possible to get the `stdin` from `flake8` though, but that isn't available in Flake8 2.x. I can provide three commits shortly:

1. The commit fixing the `stdin` issue in Flake8 2.x, but this won't work with Flake8 3.x. (xZise/flake8-quotes@cc444c5)
2. A commit using `flake8.utils.get_stdin_value` instead of `pep8.get_stdin_value` which fixes the issue mentioned in this comment.
3. A commit with the two classes mentioned in my previous comment.

Now regarding your new comment why I would want to work around pep8:

1. This is a plugin of Flake8 not pep8, so I think it shouldn't depend on pep8, especially as Flake8 3.x seems to have gotten more independent. Not sure if the long goal for Flake8 is to be independent of pep8 which is then just another plugin.
2. Apart from being a Flake8 plugin, not relying on pep8 prevents this issue in the first place. There would be no need to switch from pep8 to pycodestyle. Does pip uninstall pep8 when you update this plugin?
3. As mentioned here, we can't use it with Flake8 3.x. At least I couldn't find an easy way.
4. Additionally Flake8 3.x provides an even simpler interface (although because we need to support Flake8 2.x it isn't as simple as it could be).

twolfson:
Everything sounds good except for the 2 classes; those seem like overkill. How about something like:

```python
pep8.stdin_get_value().splitlines(True) || sys.stdin.read() || ''
```

xZise:
So I uploaded all the different commits. I disagree that the two classes are overkill, especially as they avoid to use pep8 or any stdin special logic in Flake8 3.x. Please take a look at the commit to check whether it is still overkill.

Anyway in Flake8 3.x your suggestion won't work because `sys.stdin.read()` is also empty:
```
$ echo 'f = "foobar"' | flake8 - --select=Q
''
```

There I modified `get_file_contents`:

    def get_file_contents(self):
        if self.filename in ('stdin', '-', None):
            s = sys.stdin.read()
            print(repr(s))
            return s.splitlines(True)
        else:
            return pep8.readlines(self.filename)

twolfson:
Ah, I misinterpretted "It is possible to get the stdin from flake8 though, but that isn't available in Flake8 2.x.". I thought you meant `stdin.read()` worked as expected in `Flake8@3.x` whereas we had to use `get_file_contents()` prior.

twolfson:
After reading into it a bit more, I understand why we need 2 classes; `flake8` is doing parameter sniffing. Can you open a PR so we can start commenting on the code directly? I have some feedback but it's complex to give here.

xZise:
I think this patch should work similar to tk0miya/flake8-coding@70ba7ae and still import pep8 if pycodestyle couldn't be imported (aka used on an old install where Flake8 still uses pep8).

twolfson:
Closing this issue for now as explained in https://github.com/zheller/flake8-quotes/issues/30#issuecomment-231320109

xZise:
After my initial “bashing” it would be possible to use this patch instead of #38. Especially as #38 uses `from flake8.engine import pep8`, which seems not the way to go apparently.

The main changes I'd like to see:
* Import `pep8` as a fallback
* Maybe depend on `flake8` instead of `pep8`, because then it's probably going to use the import Flake8 is using (although you could install Flake8 <2.6 and separately pycodestyle in which case it'd use pycodestyle not pep8)

And maybe the last point isn't really solvable and in which case using Flake8's import is actually the better idea.

Issue #28, opened 2016-06-01
Check prefixed strings


xZise:
Would it be sensible to check raw strings, `unicode` or `bytes`? I'm not exactly sure what your intentions are, but in at least `u'…'` should be checked, as it is a normal string on Python 3. And the change seems reasonably simple as the token contains the prefix. You could either strip the first character if it's in `rub` or if it is not in `"'`. Any invalid prefix should result in an error anyway.

You could use different error codes so that if someone doesn't care about how the quotes are formatted in raw strings, they can just ignore that code.

twolfson:
Ah, nice catch. I think a lack of support was unintentional. It looks like the fix should be adding a few more `startswith` checks (e.g. `b`, `br`, `u`, `UR`) :

```
>>> [x for x in tokenize.generate_tokens(lambda L=iter(['"abc"']): next(L))][(3, '"abc"', (1, 0), (1, 5), '"abc"'), (0, '', (2, 0), (2, 0), '')]
>>> [x for x in tokenize.generate_tokens(lambda L=iter(['u"abc"']): next(L))]
[(3, 'u"abc"', (1, 0), (1, 6), 'u"abc"'), (0, '', (2, 0), (2, 0), '')]
>>> [x for x in tokenize.generate_tokens(lambda L=iter(['r"abc"']): next(L))]
[(3, 'r"abc"', (1, 0), (1, 6), 'r"abc"'), (0, '', (2, 0), (2, 0), '')]
>>> [x for x in tokenize.generate_tokens(lambda L=iter(['b"abc"']): next(L))]
[(3, 'b"abc"', (1, 0), (1, 6), 'b"abc"'), (0, '', (2, 0), (2, 0), '')]
>>> [x for x in tokenize.generate_tokens(lambda L=iter(['ur"abc"']): next(L))]
[(3, 'ur"abc"', (1, 0), (1, 7), 'ur"abc"'), (0, '', (2, 0), (2, 0), '')]
```

https://github.com/python-git/python/blob/715a6e5035bb21ac49382772076ec4c630d6e960/Lib/tokenize.py#L103-L120

Would you be interested in submitting a PR to add support?

twolfson:
This has been resolved by @xZise in #32 and released in `0.4.0`

PR #32, opened 2016-07-04
Ignore string prefixes like "b" or "u"

xZise:
When the string uses prefixes, the checker wasn't checking those strings, because it only handles strings that start with a bad quote. To handle this, it gets the last character of the string (which is the quote character used) and removes all characters in front of the first quote character.

It also adds a few tests using one or two prefixes.

This is an implementation for #28.

### DIFF
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -77,15 +77,22 @@ class QuoteChecker(object):
                 # ignore non strings
                 continue
 
-            if not token.string.startswith(self.inline_quotes['bad_single']):
+            # Remove any prefixes in strings like `u"foo"`. Use the quote
+            # character at the end of the string to know where the prefixes end
+            quote_char = token.string[-1]  # Does only contain one character!
+            quote_start = token.string.index(quote_char)
+            text = token.string[quote_start:]
+            # `"foo"`, `b"foo"`, `br"foo"` all result in `"foo"`
+
+            if not text.startswith(self.inline_quotes['bad_single']):
                 # ignore strings that do not start with our quotes
                 continue
 
-            if token.string.startswith(self.inline_quotes['bad_multiline']):
+            if text.startswith(self.inline_quotes['bad_multiline']):
                 # ignore multiline strings
                 continue
 
-            if self.inline_quotes['good_single'] in token.string:
+            if self.inline_quotes['good_single'] in text:
                 # ignore alternate quotes wrapped in our quotes (e.g. `'` in `"it's"`)
                 continue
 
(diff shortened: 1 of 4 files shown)


flake8_quotes/__init__.py
```diff
@@ -77,15 +77,19 @@ def get_quotes_errors(self, file_contents):
                 # ignore non strings
                 continue
 
-            if not token.string.startswith(self.inline_quotes['bad_single']):
+            # Remove any prefixes in strings like u"foobar". Use the quote
+            # character at the end of the string to know where the prefixes end
+            text = token.string[token.string.index(token.string[-1]):]
```
twolfson: There is a lot going on in this line. Can we break it down into other variables to make it easier to follow?

twolfson: Also, comment examples would be great. For example:

```python
# character at the end of the string to know where the prefixes end
#   `"a"` -> `a`
#   `u"a"` -> `a`
#   `ur"a"` -> `a`
```

twolfson:
At a high level, this looks good. Left some notes on how to improve the PR for future maintenance

flake8_quotes/__init__.py
xZise: Is possible, do you want it in a separate commit or should I squash it into this one? Also a small note: Obviously this doesn't obliterate the quotes, only the prefix (e.g. the result would be always `"a"`.

twolfson: That's why I wanted the comment ;) Please add/squash it to the current commit.

xZise:
Okay squashed it into one commit and this should address your comments you had.

flake8_quotes/__init__.py
```diff
@@ -77,15 +77,22 @@ def get_quotes_errors(self, file_contents):
                 # ignore non strings
                 continue
 
-            if not token.string.startswith(self.inline_quotes['bad_single']):
+            # Remove any prefixes in strings like `u"foo"`. Use the quote
+            # character at the end of the string to know where the prefixes end
+            quote_char = token.string[-1]  # Does only contain one character!
+            quote_start = token.string.index(quote_char)
+            text = token.string[quote_start:]
+            # `"foo"`, `b"foo"`, `br"foo"` all result in `"foo"`
```
twolfson: This should likely go at the head with the rest of the comments and we should mention that it's `text's` value but to avoid wasting even more time I will land this + make the minor edits.

twolfson:
This has been merged/released in `0.4.0`. I went a little trigger happy with shuffling comments/renaming variables to be more semantic.

Issue #29, opened 2016-06-29
Incompatible with flake8 version 3


xZise:
There is a beta version of flake8 version 3, which doesn't use `config_options` anymore so I get the following error:

```
Traceback (most recent call last):
  File "/home/xzise/.pyenv/versions/3.5.1/bin/flake8", line 11, in <module>
    sys.exit(main())
  File "/home/xzise/.pyenv/versions/3.5.1/lib/python3.5/site-packages/flake8/main/cli.py", line 16, in main
    app.run(argv)
  File "/home/xzise/.pyenv/versions/3.5.1/lib/python3.5/site-packages/flake8/main/application.py", line 293, in run
    self._run(argv)
  File "/home/xzise/.pyenv/versions/3.5.1/lib/python3.5/site-packages/flake8/main/application.py", line 279, in _run
    self.initialize(argv)
  File "/home/xzise/.pyenv/versions/3.5.1/lib/python3.5/site-packages/flake8/main/application.py", line 270, in initialize
    self.register_plugin_options()
  File "/home/xzise/.pyenv/versions/3.5.1/lib/python3.5/site-packages/flake8/main/application.py", line 150, in register_plugin_options
    self.check_plugins.register_options(self.option_manager)
  File "/home/xzise/.pyenv/versions/3.5.1/lib/python3.5/site-packages/flake8/plugins/manager.py", line 451, in register_options
    list(self.manager.map(register_and_enable))
  File "/home/xzise/.pyenv/versions/3.5.1/lib/python3.5/site-packages/flake8/plugins/manager.py", line 261, in map
    yield func(self.plugins[name], *args, **kwargs)
  File "/home/xzise/.pyenv/versions/3.5.1/lib/python3.5/site-packages/flake8/plugins/manager.py", line 447, in register_and_enable
    call_register_options(plugin)
  File "/home/xzise/.pyenv/versions/3.5.1/lib/python3.5/site-packages/flake8/plugins/manager.py", line 357, in generated_function
    return method(optmanager, *args, **kwargs)
  File "/home/xzise/.pyenv/versions/3.5.1/lib/python3.5/site-packages/flake8/plugins/manager.py", line 207, in register_options
    add_options(optmanager)
  File "/home/xzise/.pyenv/versions/3.5.1/lib/python3.5/site-packages/flake8_quotes/__init__.py", line 38, in add_options
    parser.config_options.extend(['quotes', 'inline-quotes'])
```

More information:
[...]

PR #33, opened 2016-07-04
Support multiline quote check

xZise:
This also checks multiline quotes which can be configured separately. It also uses a different error code, so that repositories which don't want to use that new feature can just add the code to the ignored ones.

This is another implementation fixing #25 similar to #26 (the additional tests from there could be transferred). The main reason for this pull request is that it removes the need for the dictionaries. It also does not complain when the string contains the wanted quote also in the multiline mode (not 100% sure how the other patch handles that), by using the same idea for both variants.

There are a few things to mention:
- The messages for Q000 and Q001 are very different because I don't think that the current message “Remove bad quotes” is very descriptive. But for backwards compatibility I left the current message.
- ~~I'd prefer if #32 could be merged first. This is the reason for `text = token.string`.~~
- ~~It changes code which would need to be overhauled when support for flake8 v3 is implemented (see #29 and #34). If possible this should be merged after it (especially as fixing #29 isn't that complex).~~

### DIFF
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -11,21 +11,8 @@ class QuoteChecker(object):
     name = __name__
     version = __version__
 
-    INLINE_QUOTES = {
-        # When user wants only single quotes
-        '\'': {
-            'good_single': '\'',
-            'good_multiline': '\'\'\'',
-            'bad_single': '"',
-            'bad_multiline': '"""',
-        },
-        '"': {
-            'good_single': '"',
-            'good_multiline': '"""',
-            'bad_single': '\'',
-            'bad_multiline': '\'\'\'',
-        },
-    }
+    INLINE_QUOTES = ['"', "'"]
+    MULTILINE_QUOTES = ['"""', "'''"]
 
     def __init__(self, tree, filename='(none)', builtins=None):
         self.filename = filename
@@ -54,11 +41,15 @@ class QuoteChecker(object):
     @classmethod
     def add_options(cls, parser):
         cls._register_opt(parser, '--quotes', action='store',
-                          parse_from_config=True,
+                          parse_from_config=True, choices=cls.INLINE_QUOTES,
                           help='Deprecated alias for `--inline-quotes`')
         cls._register_opt(parser, '--inline-quotes', default='\'',
                           action='store', parse_from_config=True,
+                          choices=cls.INLINE_QUOTES,
                           help='Quote to expect in all files (default: \')')
+        cls._register_opt(parser, '--multiline-quotes', action='store',
+                          choices=cls.INLINE_QUOTES,
+                          help='Quote to expect in all files (default same as inline quote)')
 
     @classmethod
     def parse_options(cls, options):
@@ -66,9 +57,13 @@ class QuoteChecker(object):
             # https://docs.python.org/2/library/warnings.html#warnings.warn
             warnings.warn('flake8-quotes has deprecated `quotes` in favor of `inline-quotes`. '
                           'Please update your configugration')
-            cls.inline_quotes = cls.INLINE_QUOTES[options.quotes]
-        else:
-            cls.inline_quotes = cls.INLINE_QUOTES[options.inline_quotes]
+            options.inline_quotes = options.quotes
+
+        cls.inline_quotes = options.inline_quotes
+        if options.multiline_quotes is None:
+            options.multiline_quotes = options.inline_quotes
+        # assume that multiline quotes are all 3x singleline quotes
+        cls.multiline_quotes = options.multiline_quotes * 3
 
     def get_file_contents(self):
         if self.filename in ('stdin', '-', None):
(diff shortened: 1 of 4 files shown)


flake8_quotes/__init__.py
```diff
@@ -10,42 +10,40 @@ class QuoteChecker(object):
...
-    }
+    INLINE_QUOTES = set(['"', "'"])
+    MULTILINE_QUOTES = set(['"""', "'''"])
 
     def __init__(self, tree, filename='(none)', builtins=None):
         self.filename = filename
 
     @classmethod
     def add_options(cls, parser):
+        # We don't care how it is exactly sorted, just that it is deterministic
```
twolfson: This comment makes no sense to other developers. Maybe say:

```python
# Convert `cls.INLINE_QUOTES` to a deterministic iterator
```

twolfson: But honestly, why do we need a `set`? Keeping them as lists earlier on would also work and be deterministic.

twolfson:
I've started reviewing this but realizing I'm not in the best mindset. I will re-review this by the end of next weekend.

flake8_quotes/__init__.py
xZise: Now `optparse` requires the choices to be a list (`argparse` on the other hand doesn't), so I needed to convert it.

I personally prefer to use the correct iteration type (easily) available and here it is imho `set`: There is no inherent order and there are only lookup requests to it (apart from whatever `optparse` is doing) where `set` is the fastest. Granted this isn't really speed sensitive but if there is a reasonable type already there why not use it.

As you mentioned you want to re-review it next week I leave it for now and maybe just using a list is a reasonable compromise.

xZise:
That is fine, as it would allow me to solve the two dependencies. As a note, what is your take on the error messages? Should the current message be updated similar to the one here? I could also extract the choices feature into a separate patch.

twolfson:
I'm realizing that this is touching the same code as #34. Going to wait to review this until #34 is landed due to its priority

xZise:
Yeah, it is one of the points mentioned in the op.

twolfson:
Ah, sorry. I missed that x_x

xZise:
Okay I don't think this is blocked anymore. I've rebased it onto master and changed the sets into lists.

twolfson:
Removed blocked tag and re-reviewing PR

flake8_quotes/__init__.py
```diff
@@ -54,21 +41,29 @@ def _register_opt(parser, *args, **kwargs):
...
             warnings.warn('flake8-quotes has deprecated `quotes` in favor of `inline-quotes`. '
                           'Please update your configugration')
-            cls.inline_quotes = cls.INLINE_QUOTES[options.quotes]
-        else:
-            cls.inline_quotes = cls.INLINE_QUOTES[options.inline_quotes]
+            options.inline_quotes = options.quotes
+
+        cls.inline_quotes = options.inline_quotes
+        if options.multiline_quotes is None:
+            options.multiline_quotes = options.inline_quotes
```
twolfson: We discussed this in #26, we want to support backwards compatibility so we won't lint multiline strings unless there is an explicit option to lint them.

twolfson:
There was a lot of discussion inside of #26, would it be possible to commandeer that PR and fix up the remaining issues rather than restarting a new PR/discussion?

flake8_quotes/__init__.py
xZise: But here I'm using a different error code which could be ignored. Afaik there is no need to add additional overhead here.

xZise:
Would that mean overwriting their commits with mine or fixing up their commits? Because I intentionally made my own patch as I disagree with a few design choices of the original PR. And is it possible to “overtake” other people's PRs?

twolfson:
You would submit another PR with a second commit on top of the original commit. For now, let's conflate the 2 error codes as `Q000`. We can make a major release which breaks up those messages/has larger design implications.

xZise:
What do you mean with second and original commit?

xZise:
Okay anyway thinking about it, if you don't want to break other people's build the best solution would be to have it by default deactivated (like the other PR) but also use a different error code (like this one).

This wouldn't break anyone's build as it isn't enabled at first, and they can enable it when they are ready. And after a long enough wait it could be enabled by default and it would only break those who didn't enable it manually. But if it is obvious from the documentation that this is going to happen, they “have been warned” 😉.

We could use the same error code, but I don't see any reason in it, as we should change the error code at some point anyway. And that might break other user's expectations.

twolfson:
The original commit would be the commit from #26 (or at least their squashed result) and the second commit would be additional changes from you.

xZise:
Okay. Now I don't know how someone else apart from the creator can modify #26. Now someone else could create a new PR (which you mentioned) and then I rebase my changes from here onto theirs. If that is that what you suggested in [your comment](https://github.com/zheller/flake8-quotes/pull/33#issuecomment-230734711)?), I could go ahead.

xZise:
Oh I didn't mention any reason why we shouldn't use the same error code, but I think there actually might be one. If someone enables multiline quotes now, they would get `Q000` so if there is something that actually interprets that, it'll break when it get changed into `Q001`.

xZise:
Mah did press the wrong button 😑

twolfson:
This PR seems to have stalled out. I'm going to close it as it's going to conflict with #50 and needs further work to support Windows CLI as a result

PR #34, opened 2016-07-04
Support flake8 version 3

xZise:
With flake8 version 3 defining configuration variables for the configuration file has been changed and breaks with the implementation in version 2. This is using and modifying the suggested fix to support both versions.

This is fixing #29, although just running `flake8` using version 3.0.0b2 actually crashes it (see https://gitlab.com/pycqa/flake8/issues/157).


flake8_quotes/__init__.py
```diff
@@ -29,13 +30,35 @@ class QuoteChecker(object):
     def __init__(self, tree, filename='(none)', builtins=None):
         self.filename = filename
 
+    @staticmethod
+    def _register_opt(parser, *args, **kwargs):
+        """
+        Handler to register an option for both flake8 version 3 and version 2.
```
twolfson: Please use the proper casing here for "Flake8"

flake8_quotes/__init__.py
```diff
@@ -29,13 +30,35 @@ class QuoteChecker(object):
...
+
+        This is based on:
+        http://flake8.pycqa.org/en/latest/plugin-development/cross-compatibility.html
+
+        It only supports `parse_from_config` from the original function and it
+        uses the `Option` object returned to get the string.
+        """
+        try:
+            # Flake8 3.x registration
```
twolfson: We should be consistent about saying "Flake8 version 3" or "Flake8 3.x". Please choose one variant and use that consistently. My personal preference is `Flake8@3` but the current ones are good as well.

flake8_quotes/__init__.py
```diff
@@ -29,13 +30,35 @@ class QuoteChecker(object):
     def __init__(self, tree, filename='(none)', builtins=None):
         self.filename = filename
 
+    @staticmethod
+    def _register_opt(parser, *args, **kwargs):
+        """
+        Handler to register an option for both flake8 version 3 and version 2.
+
+        This is based on:
+        http://flake8.pycqa.org/en/latest/plugin-development/cross-compatibility.html
```
twolfson: This link current points to `/latest` which could easily break. Maybe we should point to a tagged GitHub document:

https://github.com/PyCQA/flake8/blob/3.0.0b2/docs/source/plugin-development/cross-compatibility.rst#option-handling-on-flake8-2-and-3

twolfson:
Implementation looks good, left some pedantic notes for comments.

xZise:
Squashed fixes into commit and rebased it onto the new master.

xZise:
Okay I have been updating the commit message to also use the `Flake8 2.x`/`Flake8 3.x` style you suggested. Maybe I could also add running tests on Flake8 3.x, although that requires (for now) a different call structure as mentioned in the OP.

flake8_quotes/__init__.py
twolfson: This comment has been resolved :+1:

twolfson:
Looks good to me. I'm going to take a shot at updating the Travis CI setup after landing this.

twolfson:
This has been landed in `0.5.0`

PR #35, opened 2016-07-05
Added test for Flake8 3.x in Travis CI

twolfson:
We now support Flake8@3 as of `0.5.0` via #34. In this PR:

- Added tests to verify Flake8 3.x is properly supported and won't break

/cc @zheller @xZise

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,17 @@ language: python
 python:
   - "2.7"
   - "3.3"
+matrix:
+  include:
+    - python: "3.3"
+      env: FLAKE8_VERSION=3
 
 install:
   # Install our dependencies
   - pip install -r requirements-dev.txt
+
+  # If we are testing Flake8 3.x, then install it
+  - if test "$FLAKE8_VERSION" = "3"; then pip install flake8==3.0.0b2; fi
 
   # Install `flake8-quotes`
   - python setup.py develop


xZise:
Would `--pre` work when installing Flake8 3.x without needing to provide a version number? Also I have written a patch addressing that in a different way in my [test-on-travis branch](https://github.com/xZise/flake8-quotes/tree/test-on-travis).

Now I haven't actually rechecked this, but afaik doesn't flake8 allow you to check empty files (see [bug 157](https://gitlab.com/pycqa/flake8/issues/157)) and `test.sh` does actually check `test/__init__.py` which is empty. I got locally errors and so I never ran it on Travis it worked there, just assumed it wouldn't work. But I can't see any error and I can't see Flake8 failing on that file.

I'll try to give it a look after I've finished the commits mentioned in #31.

twolfson:
The gists of our PRs are the same. For reference, I chose `FLAKE8_VERSION` for futureproofing of more future versions and am using `matrix.include` to avoid unnecessary excess testing (e.g. 1 more test for Python 2).

With respect to empty files, we run `flake8` inside of `test.sh` which isn't failing on Travis CI so I think we are good.

With respect to `--pre`, I'm unable to install `Flake8@3.x` locally without an explicit tag:

```bash
$ pip install --pre flake8==3.x
Collecting flake8==3.x
  Could not find a version that satisfies the requirement flake8==3.x (from versions: 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1.1, 1.2, 1.3, 1.3.1, 1.4, 1.5, 1.6, 1.6.1, 1.6.2, 1.7.0, 2.0, 2.1.0, 2.2.0, 2.2.1, 2.2.2, 2.2.3, 2.2.4, 2.2.5, 2.3.0, 2.4.0, 2.4.1, 2.5.0, 2.5.1, 2.5.2, 2.5.3, 2.5.4, 2.5.5, 2.6.0, 2.6.1, 2.6.2, 3.0.0b1, 3.0.0b2)
  No distributions matching the version for flake8==3.x
# Same result for ==3 and ==3.x.x
```

xZise:
I'm not too worried about an additional test as they run quite fast (and concurrently anyway). But just one is fine too.

What I don't understand that it doesn't fail, as I can get it reliably to fail on my system. I'll try a few things out using your patch on Travis.

xZise:
Okay I can't really figure out what is different that Travis doesn't fail.

Regarding `--pre` I had something like the following in mind:

```
$ pip install --pre flake8
Collecting flake8
  Using cached flake8-3.0.0b2-py2.py3-none-any.whl
Requirement already satisfied (use --upgrade to upgrade): pycodestyle<2.1.0,>=2.0.0 in /home/xzise/.pyenv/versions/3.5.2/lib/python3.5/site-packages (from flake8)
Requirement already satisfied (use --upgrade to upgrade): pyflakes!=1.2.0,!=1.2.1,!=1.2.2,<1.3.0,>=0.8.1 in /home/xzise/.pyenv/versions/3.5.2/lib/python3.5/site-packages (from flake8)
Requirement already satisfied (use --upgrade to upgrade): mccabe<0.6.0,>=0.5.0 in /home/xzise/.pyenv/versions/3.5.2/lib/python3.5/site-packages (from flake8)
Installing collected packages: flake8
Successfully installed flake8-3.0.0b2
```

But obviously this might install a version which is to new for your liking. The main reason I'm suggestion this is because it currently (as does my patch) only tests against 3.0.0b2, but it should keep up with new beta releases. So maybe `>=3.0.0b2;<3.0.0`?

twolfson:
I think we can submit a PR when a new beta release is out. I'm not too concerned about lagging behind by a day or 2.

xZise:
Okay regarding the issue on empty files, it was actually my plugin which cause the error, so from that standpoint it is fine what you are doing. In general I don't really have any objections.

twolfson:
Alright, going to land this :shipit:

twolfson:
This has been merged/released in `0.5.1`

PR #37, opened 2016-07-05
Support stdin in Flake8

xZise:
Use Flake8's pep8
---
As mentioned in #30 using `stdin` doesn't work anymore. Now I don't know what the original issue was, but the issue I got can be fixed by using Flake8's `pep8` package. As far as I can tell the reason behind the error is that this plugin uses the original `pep8` package while Flake8 uses `pycodestyle`. 

And when Flake8/pycodestyle reads the content from `stdin`, it gets cached to future calls return the cached value and not an empty string. But when now pep8 reads from `stdin` it reads nothing, because everything was already read from the other package.

Empty result in Flake8 3.x
---
Additionally with Flake8 3.x it seems that it doesn't use pycodestyle's implementation at all anymore. Instead Flake8 itself first reads the content and caches it. But this now causes the fix to return an empty string.

Using Flake8 3.x features to get actual content
---
To avoid this it can use the utilities provided by Flake8 itself. It'll read the parameters of the constructors and automatically provide the values “requested” and there we can request the content.

Unfortunately this doesn't work in Flake8 2.x, so it only exposes the constructor that it expects and in Flake8 2.x the plugin itself reads the content like the original patch.

Alternative ideas
---
An alternative would be to requests the tokens already and not the lines itself. Another alternative is in xZise/flake8-quotes@b983f66, but to be honest I actually don't know if I tested that version.

Also I can separate the two patches, so that it can first fix `stdin` and then actually support also Flake8 3.x (as that is not stable yet it isn't as important afaik).

**Note**: This will change the dependency from pep8 to flake8! It might not be something you want.

### DIFF
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -2,12 +2,17 @@ import optparse
 import tokenize
 import warnings
 
-from flake8.engine import pep8
+import flake8
 
 from flake8_quotes.__about__ import __version__
 
+FLAKE8_VERSION2 = int(flake8.__version__.split('.')[0]) == 2
+if FLAKE8_VERSION2:
+    # Use the pep8/pycodestyle package used by Flake8
+    from flake8.engine import pep8
 
-class QuoteChecker(object):
+
+class QuoteCheckerBase(object):
     name = __name__
     version = __version__
 
@@ -27,8 +32,8 @@ class QuoteChecker(object):
         },
     }
 
-    def __init__(self, tree, filename='(none)', builtins=None):
-        self.filename = filename
+    def __init__(self, lines):
+        self.lines = lines
 
     @staticmethod
     def _register_opt(parser, *args, **kwargs):
@@ -71,10 +76,7 @@ class QuoteChecker(object):
             cls.inline_quotes = cls.INLINE_QUOTES[options.inline_quotes]
 
     def get_file_contents(self):
-        if self.filename in ('stdin', '-', None):
-            return pep8.stdin_get_value().splitlines(True)
-        else:
-            return pep8.readlines(self.filename)
+        return self.lines
 
     def run(self):
         file_contents = self.get_file_contents()
@@ -128,6 +130,27 @@ class QuoteChecker(object):
                 'line': start_row,
                 'col': start_col,
             }
+
+
+if not FLAKE8_VERSION2:
+    class QuoteChecker(QuoteCheckerBase):
+
+        def __init__(self, tree, lines):
+            # This is using the `tree` argument as Flake8 3.0.0b2 won't run it
+            # if it doesn't contain certain parameters.
+            # See also: https://gitlab.com/pycqa/flake8/issues/159
+            super(QuoteChecker, self).__init__(lines)
+
+else:
+    class QuoteChecker(QuoteCheckerBase):
+
+        def __init__(self, tree, filename):
+            if filename in ('stdin', '-', None):
+                lines = pep8.stdin_get_value().splitlines(True)
+            else:
+                lines = pep8.readlines(filename)
+            super(QuoteChecker, self).__init__(lines)
+
 
 
 class Token:
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -2,7 +2,7 @@ import optparse
 import tokenize
 import warnings
 
-import pep8
+from flake8.engine import pep8
 
 from flake8_quotes.__about__ import __version__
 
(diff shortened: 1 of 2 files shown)


twolfson:
With respect to breaking it into 2 PRs, feel free to do that if you feel this one is moving too slow.

flake8_quotes/__init__.py
```diff
@@ -2,12 +2,17 @@
...
 
-import pep8
+import flake8
 
 from flake8_quotes.__about__ import __version__
 
+FLAKE8_VERSION2 = int(flake8.__version__.split('.')[0]) == 2
+if FLAKE8_VERSION2:
+    # Use the pep8/pycodestyle package used by Flake8
+    from flake8.engine import pep8
```
twolfson: Instead of sniffing version from `flake8.__version__`, maybe we can use `try: ... except ImportError:` to determine if the `pep8` engine exists?

flake8_quotes/__init__.py
```diff
@@ -130,6 +132,27 @@ def get_quotes_errors(self, file_contents):
             }
 
 
+if not FLAKE8_VERSION2:
+    class QuoteChecker(QuoteCheckerBase):
```
twolfson: How about we define these as `Pep8QuoteChecker` and `NoPep8QuoteChecker`? Then, we can export based on `pep8` existing or not:

```python
QuoteChecker = Pep8QuoteChecker if pep8 else NoPep8QuoteChecker
```

flake8_quotes/__init__.py
xZise: The pep8 engine exists in Flake8 3.x as well, but it isn't used so it's empty.

twolfson: What do you mean by it's empty? Can you link us?

xZise: I added a print of `file_contents` in `run` and here is what you get:

```
$ echo 'f = "foobar"' | flake8 - --select=Q
[]
$ flake8 --version
3.0.0b2 (pycodestyle: 2.0.0, mccabe: 0.5.0, pyflakes: 1.2.3, flake8_quotes: 0.5.0, flake8-future-import: 0.4.3) CPython 3.5.1 on Linux
$ git rev-parse HEAD
26f7a56a3ca1af576609c97ca42de992bc92ef82
```

This is running the first commit of this PR (so without the two class-magic). I actually mention this in the OP:

> Additionally with Flake8 3.x it seems that it doesn't use pycodestyle's implementation at all anymore. Instead Flake8 itself first reads the content and caches it. But this now causes the fix to return an empty string.

twolfson: Ah, alright. I was asking for clarification because the "it" was vague (e.g could be `pep8` or any other variable)

twolfson: This sounds like it's more of an issue in `flake8` though than us. I think you mentioned something about a corresponding issue. Maybe we should wait until that's resolved since this is a beta release?

xZise: About that I didn't mention any corresponding issue. I'm going to separate the first commit so this issue doesn't stall that part.

xZise: Okay here is the bug report about that: https://gitlab.com/pycqa/flake8/issues/161

twolfson:
Based on the response on https://gitlab.com/pycqa/flake8/issues/161, I think it would be best to wait until we get consistent `get_stdin_value` in `flake8's` beta so we can code up a simpler solution (rather than the multi-class one). Closing #37 and #38 for now.

xZise:
Okay I'd say this PR should be opened again. As it seems using `pep8` or `pycodestyle` is [not supported anymore in Flake8 3.x](https://gitlab.com/pycqa/flake8/commit/d69cb0564ada817b1cda5af49a552684466eb2dc) to get the `stdin` contents. I also worked on using `tokens` instead of `lines`, but that doesn't actually [contain all the tokens](https://gitlab.com/pycqa/flake8/issues/172).

So as supporting Flake8 3.x is not as high as a priority, I'd say reopen but wait for the possibility to get the tokens directly. Then I can rewrite this PR to use the tokens. If you can't wait that's fine too.

xZise:
Okay I should've read completely. As it seems it's possible to use the new `flake8-polyfill` package.

twolfson:
It looks like that was landed recently, I think we should hold off on changing our repo until `Flake8==3` is out of its beta phase. It seems like we could use a volatile API and waste our own time unnecessarily =/

xZise:
Well 3.0.0 got released now. Also using tokens directly isn't actually possible in the way I thought it would so I provided a merge request https://gitlab.com/pycqa/flake8/merge_requests/80. So depending on how fast you want to support it, we could wait for it being merged or use this variant or use `flake8-polyfill`.

PR #38, opened 2016-07-07
Repaired stdin support by moving from `pep8` to `flake8`

xZise:
Instead of relying on `pep8` importing itself, it can just use Flake8's `pep8` import, which will either map to `pep8` itself or to `pycodestyle` depending on what Flake8 is using. This **changes the dependency from `pep8` to `flake8`** and in theory `pep8` could be uninstalled (if not used by Flake8). As this is actually a Flake8 plugin, I don't think it is a large issue, but maybe I'm missing something so I have highlighted it here.

This is part of #37, so it will only work on Flake8 2.x. There is a bug report on [Flake8 3.x](https://gitlab.com/pycqa/flake8/issues/161), as that seems to not use `pycodestyle`'s value and thus it'll return an empty string. But as 3.x is only beta, supporting that is not as urgent as supporting 2.x.

And the remaining changes of #37 won't change anything in the handling of Flake8 2.x but just add support for 3.x, so the idea of this patch would still be required.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,7 @@ install:
   - pip install -r requirements-dev.txt
 
   # If we are testing Flake8 3.x, then install it
-  - if test "$FLAKE8_VERSION" = "3"; then
-      pip install flake8==3.0.0b2;
-    fi
+  - if test "$FLAKE8_VERSION" = "3"; then pip install flake8==3.0.0b2; fi
 
   # Install `flake8-quotes`
   - python setup.py develop
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,13 @@ install:
   - pip install -r requirements-dev.txt
 
   # If we are testing Flake8 3.x, then install it
-  - if test "$FLAKE8_VERSION" = "3"; then pip install flake8==3.0.0b2; fi
+  # Force jobs to 1 as a workaround to avoid the PicklingError
+  # see https://gitlab.com/pycqa/flake8/issues/164
+  - if test "$FLAKE8_VERSION" = "3"; then
+      pip install flake8==3.0.0b2;
+      echo "[flake8]" > tox.ini;
+      echo "jobs=1" >> tox.ini;
+    fi
 
   # Install `flake8-quotes`
   - python setup.py develop
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,12 +13,8 @@ install:
   - pip install -r requirements-dev.txt
 
   # If we are testing Flake8 3.x, then install it
-  # Force jobs to 1 as a workaround to avoid the PicklingError
-  # see https://gitlab.com/pycqa/flake8/issues/164
   - if test "$FLAKE8_VERSION" = "3"; then
       pip install flake8==3.0.0b2;
-      echo "[flake8]" > tox.ini;
-      echo "jobs=1" >> tox.ini;
     fi
 
   # Install `flake8-quotes`
--- a/test/test_checks.py
+++ b/test/test_checks.py
@@ -1,12 +1,45 @@
 from flake8_quotes import QuoteChecker
 import os
-from unittest import TestCase
+import subprocess
+from unittest import expectedFailure, TestCase
+
+
+FLAKE8_VERSION = os.environ.get('FLAKE8_VERSION', '2')
+
+
+def expectedFailureIf(condition):
+    """Only expect failure if condition applies."""
+    if condition:
+        return expectedFailure
+    else:
+        return lambda func: func
 
 
 class TestChecks(TestCase):
     def test_get_noqa_lines(self):
         checker = QuoteChecker(None, filename=get_absolute_path('data/no_qa.py'))
         self.assertEqual(checker.get_noqa_lines(checker.get_file_contents()), [2])
+
+
+class TestFlake8Stdin(TestCase):
+    @expectedFailureIf(FLAKE8_VERSION == '3')
+    def test_stdin(self):
+        """Test using stdin."""
+        filepath = get_absolute_path('data/doubles.py')
+        with open(filepath, 'rb') as f:
+            # For some reason using "--select=Q" did suppress all outputs, so
+            # the result might contain non flake_quotes related errors
+            p = subprocess.Popen(['flake8', '-'], stdin=f,
+                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            stdout, stderr = p.communicate()
+
+        stdout_lines = stdout.splitlines()
+        self.assertEqual(stderr, b'')
+        self.assertEqual(stdout_lines, [
+            b'stdin:1:25: Q000 Remove bad quotes.',
+            b'stdin:2:25: Q000 Remove bad quotes.',
+            b'stdin:3:25: Q000 Remove bad quotes.',
+        ])
 
 
 class DoublesTestChecks(TestCase):
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -2,7 +2,7 @@ import optparse
 import tokenize
 import warnings
 
-import pep8
+from flake8.engine import pep8
 
 from flake8_quotes.__about__ import __version__
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,5 @@
 [flake8]
 max-line-length = 120
+# Force jobs to 1 as a workaround to avoid the PicklingError in Flake8 3.x
+# see https://gitlab.com/pycqa/flake8/issues/164
+jobs=1
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(
     author_email='zheller@gmail.com',
     version=about['__version__'],
     install_requires=[
-        'pep8',
+        'flake8',
     ],
     url='http://github.com/zheller/flake8-quotes/',
     long_description=LONG_DESCRIPTION,


xZise:
Okay to be honest I have an issue actually testing how it should work before. I have tested an older version of Flake8 installed (2.4.0) using pep8 and not pycodestyle and it returned an empty list. And this patch didn't matter.

An alternative way could be to use a patch similar to #31 but to actually do it like in tk0miya/flake8-coding@70ba7ae. I also don't understand really why Travis failed.

twolfson:
Based on the response on https://gitlab.com/pycqa/flake8/issues/161, I think it would be best to wait until we get consistent `get_stdin_value` in `flake8's` beta so we can code up a simpler solution (rather than the multi-class one). Closing #37 and #38 for now.

xZise:
As mentioned in #30, it is unrelated to Flake8's beta and has nothing to do with the multi-class implementation from #37. In fact this patch was written independently and then wrote 49d93b9.

But it seems that the bug 161 you mention, mention the “official” way to implement this.

twolfson:
Reopening PR as discussed in https://github.com/zheller/flake8-quotes/issues/30#issuecomment-231385209

xZise:
The build failure might be due to https://gitlab.com/pycqa/flake8/issues/164 .

twolfson:
Re-running failing build but worst case scenario is we move `flake8==3.0.0b2` to a `allow_failure`

test/test_checks.py
```diff
@@ -1,6 +1,21 @@
...
+
+import flake8
+
 from flake8_quotes import QuoteChecker
 import os
-from unittest import TestCase
+from unittest import expectedFailure, TestCase
+
+
+FLAKE8_2 = getattr(flake8, '__version_info__', (2, 0)) < (3, 0)
```
twolfson: Instead of sniffing the `flake8` version, maybe we should use the environment variable we already have:

```
import os

FLAKE8_VERSION = os.env.get('FLAKE8_VERSION', 2)

# ...

@expectedFailureIf(FLAKE8_VERSION == '3')
```

test/test_checks.py
```diff
@@ -1,6 +1,21 @@
+import subprocess
+
+import flake8
```
twolfson: This repo seems to sort imports alphabetically not by first/third/local modules (e.g. `os` comes after `flake8_quotes`). Please move these new imports into the existing grouping

test/test_checks.py
```diff
@@ -9,6 +24,30 @@ def test_get_noqa_lines(self):
         self.assertEqual(checker.get_noqa_lines(checker.get_file_contents()), [2])
 
 
+class TestFlake8Stdin(TestCase):
+
```
twolfson: For consistency with other classes, please remove this extra newline

test/test_checks.py
```diff
@@ -9,6 +24,30 @@ def test_get_noqa_lines(self):
...
+        """Test using stdin."""
+        filename = get_absolute_path('data/doubles.py')
+        env = os.environ.copy()
+        env['PYTHONIOENCODING'] = 'utf8'
+        with open(filename, 'rb') as f:
+            # For some reason using "--select=Q" did suppress all outputs, so
+            # the result might contain non flake_quotes related errors
+            p = subprocess.Popen(['flake8', '-'], stdin=f, env=env,
+                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            out, err = p.communicate()
```
twolfson: Please rename `out` and `err` to `stdout` and `stderr`. At first glance, I thought `err` was an actual exception

test/test_checks.py
```diff
@@ -9,6 +24,30 @@ def test_get_noqa_lines(self):
...
+
+        out = out.decode('utf8').splitlines()
+        self.assertFalse(err)
```
twolfson: Shouldn't `stderr` equal an empty string, not false?

test/test_checks.py
```diff
@@ -9,6 +24,30 @@ def test_get_noqa_lines(self):
...
 
 
+class TestFlake8Stdin(TestCase):
+
+    @expectedFailureIf(not FLAKE8_2)
+    def test_stdin(self):
+        """Test using stdin."""
+        filename = get_absolute_path('data/doubles.py')
+        env = os.environ.copy()
+        env['PYTHONIOENCODING'] = 'utf8'
```
twolfson: What happens if we don't include 'PYTHONIOENCODING'?

test/test_checks.py
```diff
@@ -9,6 +24,30 @@ def test_get_noqa_lines(self):
         self.assertEqual(checker.get_noqa_lines(checker.get_file_contents()), [2])
 
 
+class TestFlake8Stdin(TestCase):
+
+    @expectedFailureIf(not FLAKE8_2)
+    def test_stdin(self):
+        """Test using stdin."""
+        filename = get_absolute_path('data/doubles.py')
```
twolfson: Please rename `filename` to `filepath`; `filename` would be `doubles.py` but `filepath` is the entire `/home/.../doubles.py` string

test/test_checks.py
xZise: The pipes are usually encoded depending on the encoding used by the system. Afaik can Python 3 support `sys.stderr.encoding` (and the same for `stdout`) but Python 2 does usually not. So there is no real easy way to determine it. So I just enforce it that it's always UTF8, which most OSes already use (Linux for example, which is the OS used on Travis) so removing this won't cause an error. But on afaik does Windows not use that encoding.

test/test_checks.py
xZise: Yes `assertFalse` does not check `VALUE is False` not `not VALUE`, which an empty string is as well. (to actually check if the value is False `assertIs(VALUE, False)`)

xZise:
Okay I probably figured out why this patch failed (because it required `flake8' instead of `pep8`) and maybe I have a fix for that. I'll incorporate your suggestions then as well, but I first want to verify that I found a solution.

xZise:
If wanted I could squash the third commit (c18ae98) into the first (8281749).

.travis.yml
```diff
@@ -13,7 +13,13 @@ install:
   - pip install -r requirements-dev.txt
 
   # If we are testing Flake8 3.x, then install it
-  - if test "$FLAKE8_VERSION" = "3"; then pip install flake8==3.0.0b2; fi
+  # Force jobs to 1 as a workaround to avoid the PicklingError
+  # see https://gitlab.com/pycqa/flake8/issues/164
+  - if test "$FLAKE8_VERSION" = "3"; then
+      pip install flake8==3.0.0b2;
+      echo "[flake8]" > tox.ini;
+      echo "jobs=1" >> tox.ini;
```
twolfson: Instead of automatically adding them for this build, why not add them to our permanent `setup.cfg` file?

twolfson: *them = flake8 configuration settings for `jobs`

test/test_checks.py
```diff
@@ -1,6 +1,18 @@
 from flake8_quotes import QuoteChecker
 import os
-from unittest import TestCase
+import subprocess
+from unittest import expectedFailure, TestCase
+
+
+FLAKE8_VERSION = int(os.environ.get('FLAKE8_VERSION', 2))
```
twolfson: I don't expect versions to be a number; they are usually strings (e.g. SemVer is X.Y.Z). Please remove the `int` conversion and remove the fallback to 2. I intentionally excluded a fallback so we recieve `None` by default and will use exact matches only (so we don't accidentally pass a version like `4`)

test/test_checks.py
```diff
@@ -9,6 +21,29 @@ def test_get_noqa_lines(self):
         self.assertEqual(checker.get_noqa_lines(checker.get_file_contents()), [2])
 
 
+class TestFlake8Stdin(TestCase):
+    @expectedFailureIf(FLAKE8_VERSION > 2)
```
twolfson: As mentioned in the previous comment, please use an explicit check that `FLAKE8_VERSION` is equal to 3. We don't want to accidentally pass a test that should fail.

test/test_checks.py
```diff
@@ -9,6 +21,29 @@ def test_get_noqa_lines(self):
...
+        env = os.environ.copy()
+        env['PYTHONIOENCODING'] = 'utf8'
+        with open(filepath, 'rb') as f:
+            # For some reason using "--select=Q" did suppress all outputs, so
+            # the result might contain non flake_quotes related errors
+            p = subprocess.Popen(['flake8', '-'], stdin=f, env=env,
+                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            stdout, stderr = p.communicate()
+
+        stdout = stdout.decode('utf8').splitlines()
```
twolfson: This isn't `stdout` any more; this is `stdout_lines`. Can we please rename the new variable?

```python
stdout_lines = stdout.decode('utf8').splitlines()
```

test/test_checks.py
twolfson: But we don't have anything unicode specific in this test, we should be fine to remove this as well as the `decode` call

test/test_checks.py
xZise: Ehm no, you had a fallback which was an int! `FLAKE8_VERSION = os.env.get('FLAKE8_VERSION', 2)`

xZise: Oh and a string only hides any issues. Because if it's actually using a proper version number the check below will **silently** fail, unless we do some version number parsing or are careful what version we actually use.

twolfson: Ah, sorry. I must have forgotten to add quotes to the fallback =/ We aren't using the actual semver but it's the same category of environment variables/version numbers (which are strings)

.travis.yml
```diff
@@ -13,7 +13,12 @@ install:
   - pip install -r requirements-dev.txt
 
   # If we are testing Flake8 3.x, then install it
-  - if test "$FLAKE8_VERSION" = "3"; then pip install flake8==3.0.0b2; fi
+  # Force jobs to 1 as a workaround to avoid the PicklingError
+  # see https://gitlab.com/pycqa/flake8/issues/164
+  - if test "$FLAKE8_VERSION" = "3"; then
+      pip install flake8==3.0.0b2;
+      echo "jobs=1" >> setug.cfg;
```
twolfson: I meant to add this to the actual `setup.cfg` file (as in in the PR) with the comment of course

test/test_checks.py
```diff
@@ -1,6 +1,18 @@
 from flake8_quotes import QuoteChecker
 import os
-from unittest import TestCase
+import subprocess
+from unittest import expectedFailure, TestCase
+
+
+FLAKE8_VERSION = os.environ.get('FLAKE8_VERSION', "2")
```
twolfson: This repo uses single quotes, please move to '2' not "2"

test/test_checks.py
```diff
@@ -9,6 +21,29 @@ def test_get_noqa_lines(self):
         self.assertEqual(checker.get_noqa_lines(checker.get_file_contents()), [2])
 
 
+class TestFlake8Stdin(TestCase):
+    @expectedFailureIf(FLAKE8_VERSION == "3")
```
twolfson: Another usage of double quotes when it should be single quotes

twolfson:
There was a comment on stripping utf8 encoding altogether but it seems to have been buried. Can you reply/address that?

.travis.yml
xZise: It is only needed in Flake8 3.0.0 and is a workaround. I don't see why we should cripple the normal usage of Flake8 just because of the bug. Also 3.x is still in beta so as soon as the bug has been fixed we can remove this line again. And as it'll be hopefully fixed before it goes stable, we don't have to care about the existence of the bug.

test/test_checks.py
xZise: That is the reason the Travis tests failed.

.travis.yml
xZise: Oh and as it the Travis tests now again failed I assume that using `setup.cfg` doesn't actually work properly.

test/test_checks.py
xZise: Well I dislike undefined behavior, but it should be in theory possible to just work on bytes instead of strings.

twolfson:
We keep on losing threads due to shifting commits. With respect to the `jobs=1` setting, it will only affect this repo and I think that's a fine loss if it's well documented (via a comment) so we can remove it later on

twolfson:
With respect to encoding (same buried comments frustration), I don't understand what you mean by undefined behavior; we know the cases (Python 2 and Python 3), in Python 2 `stdout` is a "string", in Python 3 `stdout` is "bytes".

We should be able to use something like `str(stdout)` to coerce it consistently to a string so we can do `splitlines`. The point I'm trying to make is we don't need all the handwaving for UTF8 as the content is valid ASCII.

xZise:
* `setup.cfg` modification work, also it's a Travis related patch so why bother to affect more than needed?

Regarding `stdout`: In Python 2, `bytes` == `str` and thus I handle it. If it's text it's `unicode` (or `str` in Python 3) and if it's data it's `bytes` (or `str` in Python 2). Now luckily `bytes.splitlines` works so no worries and we can just compare against the `bytes` strings as I did in the patch. This is also why I initially decoded it because I treated the output as text. And it doesn't matter if the text is ASCII or non-ASCII because that just messes everything up when you say “well if we are lucky and it's only ASCII we can treat our data as if it were text”. This is by the way why I love that Python 3 strictly separates data and text. But hey we can compare data blobs against each other so there is no need to decode the output.

Also no pleas don't do `str(stdout)`? There is basically only one valid reason afaik, and that is the keys of environment variables which need to be `bytes` in Python 2 and `str` in Python 3 (at least in some cases) and there `str(…)` works reasonably well. But if apart from that I don't see a reason why we would want `bytes` in Python 2 and `str` in Python 3 as both are different type of things. And while it could work it is afaik just bad style to hope it does. Just make sure you know what you have and if you are sure you can predictably work with it: We now get a `bytes` object (although Python 2 says `str` it's just easier to think it as `bytes` as that is how Python 3 treats it) and we can split it into lines and compare each `bytes` in the list with another list of `bytes`. Now the only way this will fail if there are non-ASCII characters and the encoding of the file and the encoding of the pipe are different.

Now if there were non-ASCII my initial additions would solve that issue: We get a `bytes` object and decode it with the encoding we enforced and then compare it to a list of `str` (when I say `str` I mean `unicode` in Python 2), which are automatically decoded using the encoding of the file.

And maybe I should just provide new commits and when you are satisfied I just squash it into sensible commits (or the commits you suggested).

twolfson:
With respect to `setup.cfg`, Travis CI configurations should almost never have bespoke scripts in them (e.g. adjusting configurations). The case where including a Travis CI specific script fails in this scenario is local testing of `flake8==3.0.0b2` will fail due to the same pickling error.

xZise:
Okay but for some reason it doesn't actually read the `setup.cfg` otherwise the build wouldn't have failed: https://travis-ci.org/xZise/flake8-quotes/jobs/145271849

twolfson:
Can we research why that is? `tox.ini` seems strange since `tox` is for a specific tool which we aren't even using here =/

xZise:
Afaik Flake8 does search for `tox.ini` regardless of `tox` as it doesn't need to do anything special with it apart from just reading it as a config file and searching for the `[flake8]` section.

But it seems that it didn't actually did the right thing to the `setup.cfg`. I edited the file instead of using the `.travis.yml` and there [it worked](https://travis-ci.org/xZise/flake8-quotes/builds/145330766) (fd3da7e).

Should I squash that change into the current one or are there other changes you want me to do? Before I obliterate the commits again.

xZise:
So if you are happy with the changes I'd like to squash fd3da7e and 34d3b40 into 4d89054 so that this PR doesn't break any `git blame` in `.travis.yml` as that is unnecessary.

twolfson:
A squash won't be necessary; GitHub now supports squashing in the pull requests :+1: 

https://github.com/blog/2141-squash-your-commits

twolfson:
This looks good to me, landing this PR

twolfson:
This has been merged/squashed/released in `0.6.0`. Thanks for bearing with us and patching this bug =)

xZise:
Well the commit message is not the prettiest 😉. But thanks for bearing with me too.

xZise:
I just noticed that `flake8.engine` doesn't exist in Flake8 3.x anymore. So this has completely broken support for 3.x. Not sure what we want to do about it, but Flake8 3.x isn't officially supported by this plugin anyway. As stdin is broken anyway in Flake8 3.x, the easiest way is probably to wrap the import into a `try-except` and set `pep8` to `None` or so when `flake8.engine.pep8` couldn't be imported.

twolfson:
I concur with try/except being the simplest solution for now. Going to open a PR

twolfson:
Actually we use `pep8` for both `stdin` and reading file contents so that solution won't work. Going to reopen #30

xZise:
Well this only applies to Flake8 3.x but still fixes the 2.x issue. So using `try-except` would not change how 2.x behaves, but it would support 3.x when stdin is not used. And as mentioned in #37, Flake8 3.x won't support using `pep8`/`pycodestyle` so for that another fix is required anyway.

PR #39, opened 2016-07-08
Test using stdin as input

xZise:
As you requested in #37 these are the tests in preparation for #38. Now in general I think this is okay so, the only problematic part could be that (as explained in the comment) selecting only errors from this plugin didn't work for some reason.

Now there shouldn't be any other errors but if you want I could iterate through the list and manually filter it. That would probably require a regex or similar, to make sure it filtered correctly. Alternatively I could made the expectation and output lines a set and check whether the output is a super set. But that could ignore additional errors by this plugin and which aren't expected.

That reason in general is why I haven't done anything as we should be able to respond if we need to. And at the moment it is expected to fail anyway.

### DIFF
--- a/test/test_checks.py
+++ b/test/test_checks.py
@@ -1,12 +1,38 @@
+import subprocess
+
 from flake8_quotes import QuoteChecker
 import os
-from unittest import TestCase
+from unittest import expectedFailure, TestCase
 
 
 class TestChecks(TestCase):
     def test_get_noqa_lines(self):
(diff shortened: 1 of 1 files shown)


twolfson:
These tests belong with the corresponding code that they patch. Please relocate the tests into #38 (which I believe is to patch stdin)

twolfson:
This can be done via `git cherry-pick <branch/commit>`

xZise:
The reason I suggested this PR separately is that it can verify the changes in #38. But if you want I can incorporate them into there.

PR #42, opened 2016-07-17
Require valid parameter choice

xZise:
The `--quotes` and `--inline-quotes` options require a value of a set of certain choices.

This is basically extracted from #33, but uses the dict supplied instead. I originally couldn't extract it because #33 would introduce lists instead of dicts, so calling `list()` wouldn't be necessary and so it would unnecessarily break `git blame`. But the discussion in #26 showed that it might be sensible to use dicts to map the chosen value to the actual used value. While this is not necessary at the moment for inline quotes, the multiline quotes profit from that.

Unfortunately I was unable to write a test as `optparse` by default doesn't actually verify the choice. So I'd have to manually call the check for each value. Maybe it is possible to use `flake8` somehow but for now this should suffice.

### DIFF
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -54,10 +54,12 @@ class QuoteChecker(object):
     @classmethod
     def add_options(cls, parser):
         cls._register_opt(parser, '--quotes', action='store',
-                          parse_from_config=True,
+                          parse_from_config=True, type='choice',
+                          choices=list(cls.INLINE_QUOTES),
                           help='Deprecated alias for `--inline-quotes`')
         cls._register_opt(parser, '--inline-quotes', default='\'',
                           action='store', parse_from_config=True,
+                          type='choice', choices=list(cls.INLINE_QUOTES),
                           help='Quote to expect in all files (default: \')')
 
     @classmethod
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -55,11 +55,11 @@ class QuoteChecker(object):
     def add_options(cls, parser):
         cls._register_opt(parser, '--quotes', action='store',
                           parse_from_config=True, type='choice',
-                          choices=list(cls.INLINE_QUOTES),
+                          choices=sorted(cls.INLINE_QUOTES.keys()),
                           help='Deprecated alias for `--inline-quotes`')
         cls._register_opt(parser, '--inline-quotes', default='\'',
-                          action='store', parse_from_config=True,
-                          type='choice', choices=list(cls.INLINE_QUOTES),
+                          action='store', parse_from_config=True, type='choice',
+                          choices=sorted(cls.INLINE_QUOTES.keys()),
                           help='Quote to expect in all files (default: \')')
 
     @classmethod


flake8_quotes/__init__.py
```diff
@@ -54,10 +54,12 @@ def _register_opt(parser, *args, **kwargs):
     @classmethod
     def add_options(cls, parser):
         cls._register_opt(parser, '--quotes', action='store',
-                          parse_from_config=True,
+                          parse_from_config=True, type='choice',
+                          choices=list(cls.INLINE_QUOTES),
```
twolfson: As a programmer, I don't exactly know what happens when I convert a dictionary to a list; do I get the keys, do I get the values. Can we move to `cls.INLINE_QUOTES.keys()` here (or `sorted(cls.INLINE_QUOTES.keys())` if you want to guarantee consistent ordering as you mentioned in #33)

xZise: To be honest if someone doesn't know if it uses keys, values or both, they probably don't know what it actually does even if there is `.keys()`? I mean everybody probably at some point did iterate over a dict and only got the keys.

But sure it could explicitly call `.keys()`  and maybe even do a sorted thing.

twolfson: That's what I'm saying. A novice programmer who would like to contribute to this repo might not understand the behavior whereas calling `.keys()` is much more clear

twolfson:
We really don't need tests on option parsing; that's testing a component out of our scope (i.e. testing `argparse`). Please remove those tests as it add a maintenance tax (e.g. we would need to update the tests any time we edit an option, more content for a developer to read through).

twolfson:
The PR looks good now with https://github.com/zheller/flake8-quotes/pull/42/commits/5ae8d3f2b4f286afdfc172ba58a6669f0bc7993c (except for the tests from https://github.com/zheller/flake8-quotes/pull/42/commits/c7d62ef53826ae5cdb509abf1ad54b980b77c373 )

xZise:
There you go. I mentioned in the OP that there are no tests but today I figured out how Flake8 2.6 does it, so I thought maybe I could add tests. But yes I too thought they were more testing `optparse` than this plugin.

twolfson:
For personal reference, the `--help` dialog doesn't change but feedback on invalid input goes from:

```
$ flake8 --inline-quotes a setup.py 
...
  File "/home/todd/github/flake8-quotes/flake8_quotes/__init__.py", line 71, in parse_options
    cls.inline_quotes = cls.INLINE_QUOTES[options.inline_quotes]
KeyError: 'a'
```

to

```
$ flake8 --inline-quotes a setup.py 
Usage: flake8 [options] input ...

flake8: error: option --inline-quotes: invalid choice: 'a' (choose from '"', "'")
```

twolfson:
Looks good to me, landing this now

twolfson:
This has been merged/released in `0.7.0`. Thanks for the PR!

xZise:
I'm afraid this breaks it as long as a configuration file for it is used, unless PyCQA/pycodestyle#561 is fixed. See also [Flake8 but 176](https://gitlab.com/pycqa/flake8/issues/176) for the original report.

Not sure if you should revert it and in a later patch it could be updated together with a newer required `pycodestyle` installation. Although I have no idea how to verify that.

twolfson:
It looks like this was resolved in an upstream issue. Are we still having this problem?

https://github.com/PyCQA/pycodestyle/pull/562

xZise:
A version with that change wasn't released yet and I don't know how often they are going to release updates.

PR #44, opened 2016-07-26
Use flake8-polyfill to support Flake8 2.x and 3.x

xZise:
As Flake8 2.6 switched from `pep8` to `pycodestyle` so that using `pep8` itself doesn't help. So with 09198d1e the plugin imported `flake8.engine.pep8` which is either one of them depending on the version. But Flake8 3.x doesn't have this package anymore so it broke support for Flake8 3.x then.

This is redesigning the import to use `flake8-polyfill` which provides an easily handleable version number. So instead of using `flake8.engine.pep8` it is comparing the version number and based on that either imports `pep8` or `pycodestyle`. But additionally to that it needs to monkey patch the stdin function in Flake8 3.x into that package.

It enables the test using Flake8's stdin functionality on Flake8 3.x. It also adds a test to use Flake8 on a file.

### DIFF
--- a/test/test_checks.py
+++ b/test/test_checks.py
@@ -1,10 +1,8 @@
+from flake8_polyfill.version import version_info
 from flake8_quotes import QuoteChecker
 import os
 import subprocess
 from unittest import expectedFailure, TestCase
-
-
-FLAKE8_VERSION = os.environ.get('FLAKE8_VERSION', '2')
 
 
 def expectedFailureIf(condition):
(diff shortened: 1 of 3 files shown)


test/test_checks.py
```diff
@@ -1,12 +1,10 @@
+from flake8_polyfill.version import version_info
 from flake8_quotes import QuoteChecker
 import os
 import subprocess
 from unittest import expectedFailure, TestCase
 
 
-FLAKE8_VERSION = os.environ.get('FLAKE8_VERSION', '2')
```
xZise: I've changed this because when I locally use Flake8 3.x this variable might not be set. As we require `flake8_polyfill` anyway I figured it's okay to use it to get a replacement.

xZise:
This is another try on #29.

twolfson:
Oh, `flake8==3` is using `pycodestyle`, even for retrieving lines? I thought it was only for `stdin`

test/test_checks.py
```diff
@@ -21,24 +19,53 @@ def test_get_noqa_lines(self):
...
+        stdout, stderr = p.communicate()
+
+        stdout_lines = stdout.splitlines()
+        self.assertEqual(stderr, b'')
+        # Encode the path as ASCII and hope for the best
+        encoded_path = filepath.encode('ascii')
+        self.assertEqual(stdout_lines, [
+            encoded_path + b':1:' + self.expected_error_suffix,
+            encoded_path + b':2:' + self.expected_error_suffix,
+            encoded_path + b':3:' + self.expected_error_suffix,
```
twolfson: I have a personal philosophy about tests: if there's something dynamic going on in the "expected" half of the assertion, then we are probably doing something wrong. Humans can easily make a mistake on future code updates.

Instead of this, maybe we assert there are 3 lines and then use regexp matching for line numbers (e.g. `\d{2}`)

test/test_checks.py
```diff
@@ -21,24 +19,53 @@ def test_get_noqa_lines(self):
...
+        else:
+            cls.expected_error_suffix = b'24: Q000 Remove bad quotes.'
+            # Conversely Flake8 3.x currently has a bug that it doesn't report
+            # any error unless selected specifically
+            # See also: https://gitlab.com/pycqa/flake8/issues/183
+            parameters = ['--select=Q']
+        cls.command = ['flake8'] + parameters
+
+    def test_file(self):
+        """Test using stdin."""
```
twolfson: This docstring comment isn't correct =/

test/test_checks.py
```diff
@@ -21,24 +19,53 @@ def test_get_noqa_lines(self):
...
+        filepath = get_absolute_path('data/doubles.py')
+        p = subprocess.Popen(self.command + [filepath],
+                             stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
twolfson: We already have a file checker, it's every other test in this file =/

test/test_checks.py
xZise: I considered it to dynamic which is why I choose this path. But maybe regex is better.

test/test_checks.py
xZise: Well yes, but they don't use Flake8. And maybe Flake8 broke something.

xZise:
Regarding Flake8 3.x using `pycodestyle.readlines`: I'm actually to lazy to check, but it doesn't matter. The reason it is so messy is because Flake8/pycodestyle did read everything from `stdin` so that the plugins didn't have anything left over. So Flake8/pycodestyle provided a common way, but depending on the situation Flake8 or pycodestyle or pep8 might read stdin at first and then we need to ask the correct … thingy.

But for reading files there is no such issue and we can do whatever we like. We could even open the file manually (but we would need to decode it manually if we want to work on proper strings not bytes).

test/test_checks.py
twolfson: Right, but we aren't testing `flake8`; we are testing `flake8-quotes` ;)

twolfson:
With respect to `readlines`, it looks like it's working so I guess it works :tada:

twolfson:
This has been superceded by #46 to avoid unnecessary changes and lots of back/forth in the PR (we seem to have a lot of that =/)

test/test_checks.py
xZise: Exactly: This is testing `flake8-quotes` in Flake8. If this test existed sooner we might have seen sooner that we broke support for Flake8 3.x. But we didn't noticed it, because the only Flake8 related test was expected to fail. So I thought that only Flake8 3.x using stdin was broken wen in fact Flake8 3.x was broken completely.

Now you could argue that the stdin test is no expected to succeed with your patch, but I'd argue against that because it is only testing stdin. The normal way could still be broken.

I don't specifically test that it can detect the one file correctly. It is just one randomly chosen file, but it tests that the plugin correctly works in Flake8.

And in theory it could be expanded to actually test more things that might break, like applying a setting (so we test that `--quotes`/`--inline-quotes` are actually applied) or selecting/ignoring the error code specifically (now that we made the plugin actually compatible) or using a configuration file (which might have shown us that using `type="choice"` was breaking something).

I'm fine with your #46 patch but I personally think that this test should be added (albeit improved to be not as “messy”).

twolfson: Right but if `flake8` breaks, then it's likely a problem out of our reach and implementing a workaround is likely silly. This level of testing is usually only if we encounter a regression, otherwise we have verified it works manually so it's good enough for now.

PR #45, opened 2016-07-26
Use proper entry point name

xZise:
The entry point must be of the form `errorcode = target` so that it is possible to select the plugin explicitly in Flake8 2.x and implicitly in Flake8 3.x.

The following commands would've yielded no errors from this plugin without the change. In Flake8 2.x it wasn't possible to select it:

    flake8 --select=Q FILES

In Flake8 3.x (if the code has been made compatible again) it wasn't selected by default without the user explicitly selecting it:

    flake8 FILES

Calling the command in the other Flake8 version would have worked. See also https://gitlab.com/pycqa/flake8/issues/183

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ setup(
     include_package_data=True,
     entry_points={
         'flake8.extension': [
-            'flake8_quotes = flake8_quotes:QuoteChecker',
+            'Q000 = flake8_quotes:QuoteChecker',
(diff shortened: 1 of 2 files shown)


twolfson:
Looks good to me, landing now

twolfson:
This has been merged/released in `0.8.0`. Thanks for the fix!

PR #46, opened 2016-07-26
Repaired Flake8==3 support

### LINKED ISSUE #29: Incompatible with flake8 version 3

twolfson:
This PR repairs #29 and supercedes #44 by using less lines of code and reducing overall noise. In this PR:

- Moved to installing `flake8==3` in Travis CI
- Moved to `try/except` for loading `pep8` and `pycodestyle`
    - Attribution to @xZise in #44 for pointing out that we can use `pycodestyle` as an alternate to reading files
- Fixed #29 

/cc @zheller @xZise

### DIFF
--- a/test/test_checks.py
+++ b/test/test_checks.py
@@ -1,18 +1,7 @@
 from flake8_quotes import QuoteChecker
 import os
 import subprocess
-from unittest import expectedFailure, TestCase
-
-
-FLAKE8_VERSION = os.environ.get('FLAKE8_VERSION', '2')
-
-
-def expectedFailureIf(condition):
-    """Only expect failure if condition applies."""
-    if condition:
-        return expectedFailure
-    else:
-        return lambda func: func
+from unittest import TestCase
 
 
 class TestChecks(TestCase):
@@ -22,7 +11,6 @@ class TestChecks(TestCase):
 
 
 class TestFlake8Stdin(TestCase):
-    @expectedFailureIf(FLAKE8_VERSION == '3')
     def test_stdin(self):
         """Test using stdin."""
         filepath = get_absolute_path('data/doubles.py')
@@ -33,11 +21,10 @@ class TestFlake8Stdin(TestCase):
 
         stdout_lines = stdout.splitlines()
         self.assertEqual(stderr, b'')
-        self.assertEqual(stdout_lines, [
-            b'stdin:1:25: Q000 Remove bad quotes.',
-            b'stdin:2:25: Q000 Remove bad quotes.',
-            b'stdin:3:25: Q000 Remove bad quotes.',
-        ])
+        self.assertEqual(len(stdout_lines), 3)
+        self.assertRegexpMatches(stdout_lines[0], b'stdin:1:(24|25): Q000 Remove bad quotes.')
+        self.assertRegexpMatches(stdout_lines[1], b'stdin:2:(24|25): Q000 Remove bad quotes.')
+        self.assertRegexpMatches(stdout_lines[2], b'stdin:3:(24|25): Q000 Remove bad quotes.')
 
 
 class DoublesTestChecks(TestCase):
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ install:
   - pip install -r requirements-dev.txt
 
   # If we are testing Flake8 3.x, then install it
-  - if test "$FLAKE8_VERSION" = "3"; then pip install flake8==3.0.0b2; fi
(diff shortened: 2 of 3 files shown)


xZise:
If this is actually fixing the Flake8 3.x issues it should also remove the `expectedFailure`.

test/test_checks.py
```diff
@@ -33,11 +33,10 @@ def test_stdin(self):
 
         stdout_lines = stdout.splitlines()
         self.assertEqual(stderr, b'')
-        self.assertEqual(stdout_lines, [
-            b'stdin:1:25: Q000 Remove bad quotes.',
-            b'stdin:2:25: Q000 Remove bad quotes.',
-            b'stdin:3:25: Q000 Remove bad quotes.',
-        ])
+        self.assertEqual(len(stdout_lines), 3)
```
xZise: I think it would be better if there isn't a separate line count check before. If the output is different it might just tell the difference in line numbers, which is not really helpful. So I'd do this last, but then you might not know how the other lines looked when it fails with the first line. I wouldn't be concerned that it might compare a line that doesn't exist, because that indirectly is the same as this check (with the only difference that this check also fails when the number of lines is to high).

Not sure what is the best way here.

twolfson: Right, we want to eagerly fail if we have too many lines. Additionally, an `IndexError` might exist if we have too low of a count and do this at the end. In my experience, this is the best setup

twolfson:
Ah, good catch on `expectedFailure`. Let's see what happens.

twolfson:
This was passing locally without `FLAKE8_VERSION` so we should be good. Going to wait until Travis CI is green before landing though.

However, based on @xZise I'm taking that as an implicit "lgtm"

twolfson:
Travis CI approves, landing this now...

twolfson:
This has been merged/released in`0.8.1`

xZise:
Good (?) news: With Flake8 3.0.2 the indexing was “fixed/reverted”, so we could use the old system which didn't assumed that the column is always 25 and install 3.0.2.

Issue #49, opened 2017-01-31
Error when using the CLI option


art1415926535:
I try use the CLI option but get error.
```
> flake8 --quotes='"'
Usage: flake8 [options] file file ...

flake8: error: option --quotes: invalid choice: "''" (choose from '"', "'")
```
Without the option it works fine.

flake8-quotes (0.8.1)
Python 3.5
Windows 10

twolfson:
Windows' `cmd.exe` doesn't seem to interpret neither single nor double quotes:

```
C:\Users\IEUser>echo 'hi"'
'hi"'

C:\Users\IEUser>echo "hi'"
"hi'"
```

Can you try `flake8 --quotes="`?

Arguably, choosing to use the literals was a bad call due to escaping issues =/ Maybe we should either alias them as `--quotes=double` or remove the literals entirely?

art1415926535:
Yes, the same error.
```
flake8 --quotes="
Usage: flake8 [options] file file ...

flake8: error: option --quotes: invalid choice: '' (choose from '"', "'")
```

twolfson:
Ah, damn. Alright, I'm going to whip up a PR with aliasing support for `--quotes=double` and `--quotes=single`. Sorry for the inconvenience and thanks for bug report and timely responses =)

twolfson:
We have opened up the PR https://github.com/zheller/flake8-quotes/pull/50 to resolve the issue

twolfson:
We have landed and released #50 in `0.9.0`. Thanks again for the bug report and timely responses

PR #50, opened 2017-01-31
Added alias support for Windows CLI support

twolfson:
In #49, we received a report of Windows not supporting our CLI options. After some triage, we confirmed that this was true and there are no plausible workarounds. To remedy this, we think adding aliases for `inline-quotes` is a reasonable idea (e.g. `--inline-quotes double`). In this PR:

- Added aliases for `inline-quotes` options
- Added tests
- Added documentation

/cc @zheller


zheller:
checks out ship it

PR #52, opened 2017-05-18
Fixes for issue #25

batlock666:
I was testing flake8-quotes, and I noticed that it didn't work for multiline strings. Went ahead and fixed it, and added an option to configure multiline strings separately (see issue #25). Tests and Readme are also updated.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -28,14 +28,16 @@ By default, we expect single quotes (') and look for unwanted double quotes (").
 
 .. code:: shell
 
-    flake8 --inline-quotes '"'
+    flake8 --inline-quotes '"' --multiline-quotes "'"
     # We also support "double" and "single"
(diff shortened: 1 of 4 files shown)


twolfson:
This looks great! There are a bunch of one-offs I want to tweak (e.g. not listing `--multiline-quotes` in the initial example, adding comments for clarity, fixing tests) so I'll do that in a dependent PR

twolfson:
This has been landed/released in `0.10.0`

cas--:
This has just broken the double-quote conventions for docstrings and caused my ci tests to fail...

https://www.python.org/dev/peps/pep-0008/
> For triple-quoted strings, always use double quote characters to be consistent with the docstring convention in PEP 257 . 
https://www.python.org/dev/peps/pep-0257/

PR #53, opened 2017-05-19
Touch up for #52

twolfson:
@batlock666 did a great job in #52 and I didn't want to waste their time with my nitpicking. As a result, I've written a few commits on top of their PR. In this PR:

- Lands #52 
- Resets documentation to not include `--multiline-quotes` in the initial example
    - In my experience on other repos, people will almost always grab the default config you show even when you warn them
- Renamed `cls.inline_quotes` to `cls.config` for clarity
- Added a lot of comments/examples to the code (always dreaded reading the code prior to this because it gets confusing quick)
- Upgraded `flake8` and fixed linting error which was causing #52 to fail

/cc @batlock666 @zheller (I think I'll land with either approval)


requirements-dev.txt
```diff
@@ -1,2 +1,2 @@
-flake8<2.6.0,>=2.5.0
+flake8<3.4.0,>=3.3.0
```
zheller: does this bump affect users who just install the plugin?

twolfson: Nope, it only affects people who develop this plugin

twolfson:
Alright, shipping it :shipit:

twolfson:
This has been landed/released in `0.10.0`

PR #54, opened 2017-05-19
Corrected default setup for multiline quotes

twolfson:
After releasing, I realized that we had broken our default config so we probably broke other's too. In this PR:

- Corrected default setup for multiline quotes to be """
- Updated tests to use appropriate `0.10.0` legacy  for multiline tests
- Updated CLI description to use """


twolfson:
Going to cowboy land this as a patch due to high frequency of messages in #55. I would appreciate a second pair of eyes via @batlock666 and @zheller though

twolfson:
This has been merged/released in `0.10.1`

PR #57, opened 2017-05-23
Naively implement Q001 for multiline strings

karamanolev:
This is very naive implementation, but I think it's a good first step. A little bit of change to the logic, but it makes a bit more sense now I think. It passes all the existing tests.
What I'm planning to do next is to replace the tokenize library with the ast module, as this will allow us to recognize docstrings and generally give us more flexibility in the checks. I still don't know how that will play with py2.
Not sure if I need to add that to entry_points in setup.py

### DIFF
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -158,39 +158,26 @@ class QuoteChecker(object):
             last_quote_char = token.string[-1]
             first_quote_index = token.string.index(last_quote_char)
             unprefixed_string = token.string[first_quote_index:]
-            is_multiline_string = '\n' in unprefixed_string
+            is_multiquote_string = unprefixed_string[0] * 2 == unprefixed_string[1:3]
             start_row, start_col = token.start
 
-            if is_multiline_string:
+            if is_multiquote_string:
                 # If our string has a valid multiline start, then ignore it
-                #   (""")foo""" -> good (continue)
-                #   (''')foo' -> possibly bad
-                #   (")foo" -> possibly bad
-                #   (')foo' -> possibly bad
                 if unprefixed_string.startswith(self.config['good_multiline']):
                     continue
 
-                # Report the error
                 yield {
                     'message': 'Q001 Remove bad quotes from multiline string.',
                     'line': start_row,
                     'col': start_col,
                 }
             else:
-                # If our string is a known good string, then ignore it
-                #   (')foo' -> good (continue)
-                #   (")foo" -> possibly bad
-                if unprefixed_string.startswith(self.config['good_single']):
-                    continue
-
-                # If our string is wrapping a known good string, then ignore it
-                #   "it(')s" -> good (continue)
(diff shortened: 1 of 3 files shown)


flake8_quotes/__init__.py
```diff
@@ -158,20 +158,25 @@ def get_quotes_errors(self, file_contents):
             last_quote_char = token.string[-1]
             first_quote_index = token.string.index(last_quote_char)
             unprefixed_string = token.string[first_quote_index:]
+            is_multiline_string = '\n' in unprefixed_string
```
twolfson: That's a nice thought but it won't work for multiquote strings that are only 1 line long (e.g. `"""foo"""`). How about this logic?

```python
# (" * 3 -> """) == (""")foo"""
# (" * 3 -> """) == ("fo)o"
unprefixed_string[0] * 3 == unprefixed_string[0:2]
```

flake8_quotes/__init__.py
```diff
@@ -158,20 +158,25 @@ def get_quotes_errors(self, file_contents):
...
-            #   (")foo" -> possibly bad
-            #   (')foo' -> possibly bad
-            if unprefixed_string.startswith(self.config['good_multiline']):
-                continue
+            if is_multiline_string:
+                # If our string has a valid multiline start, then ignore it
+                #   (""")foo""" -> good (continue)
+                #   (''')foo' -> possibly bad
+                #   (")foo" -> possibly bad
+                #   (')foo' -> possibly bad
```
twolfson: We can probably ditch these inline string comments

twolfson:
This looks great for a first pass! It looks like Travis CI is failing but that's likely due to the multiline detection logic (left a comment about how to fix it).

With respect to the error codes, you're right -- we should update our `Q000` to `Q0` so we reserve the `Q0XX` space:

http://flake8.pycqa.org/en/latest/plugin-development/registering-plugins.html

karamanolev:
@twolfson I think the comments are addressed. Travis is now passing. PTAL.

flake8_quotes/__init__.py
```diff
@@ -158,42 +158,31 @@ def get_quotes_errors(self, file_contents):
...
-                #   "foo" -> possibly bad
+                yield {
+                    'message': 'Q001 Remove bad quotes from multiline string.',
+                    'line': start_row,
+                    'col': start_col,
+                }
+            else:
+                # This covers two cases:
+                # - the string starts and end with the good single quote e.g. 'asdf'
+                # - the string contains a good single quote e.g. "asdf's", which is also OK
```
twolfson: As a personal preference, I've moved back to the example-based comments. I feel they communicate points more effectively -- in past iterations on this project, the verbal comments only confuse me =/

twolfson:
This looks more/less good. I've opened #58 to pull it across the finish line

twolfson:
This has been landed/released in `0.11.0`

PR #58, opened 2017-05-23
Additions to #57

twolfson:
PR #57 looks more/less good but I have some one-offs that were going to bother me. In this PR:

- Commits from #57 
- Moved to `[0] * 3 == [0:3]` check for being straightforward
- Added back example-based comments
- Added support and tests for ignoring bad multiline string when it contains a good multiline string
- Renamed `multiquote` to `multiline` for consistency

/cc @karamanolev @zheller

### DIFF
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -158,26 +158,39 @@ class QuoteChecker(object):
             last_quote_char = token.string[-1]
             first_quote_index = token.string.index(last_quote_char)
             unprefixed_string = token.string[first_quote_index:]
-            is_multiquote_string = unprefixed_string[0] * 2 == unprefixed_string[1:3]
+
+            # Determine if our string is multiline-based
+            #   "foo"[0] * 3 = " * 3 = """
+            #   "foo"[0:3] = "fo
+            #   """foo"""[0:3] = """
+            is_multiline_string = unprefixed_string[0] * 3 == unprefixed_string[0:3]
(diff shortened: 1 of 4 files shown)


karamanolev:
Looks good. I think with this implementation, multiline is a bit of a misnomer - they might very well be single line and we would still call them multiline. In the regular Python documentation, I just checked, strings are being called single-quoted, double-quoted and triple-quoted. I agree if that's also not making it very clear. In the lexical reference (https://docs.python.org/3/reference/lexical_analysis.html#literals) they're being called short strings and long strings ... not perfect. I guess we can keep calling it multiline if a bit confusing, but triple quoted is a bit better IMO.

As for the example-based comments - I removed them because of possible misunderstanding your comment "We can probably ditch these inline string comments". I'm fine either way.

The [0] * 3 == [0:3] checks looks a bit like a == a == b == c instead of a == b == c. I actually considered doing just [0] == [1] == [2] but it was too long for one line so it had to go on 2, which would've been ugly.

twolfson:
I concur that I could have done a better word choice than inline and multline but it's what we're stuck with for now unless we want to do a breaking major release. For reference, we've had inline for a year now:

https://github.com/zheller/flake8-quotes/tree/0.3.0

I updated the code references to `multiline` for consistency with the existing CLI flags.

Going to land this PR as a minor release

twolfson:
This has been landed/released in `0.11.0`

PR #59, opened 2017-05-25
Docstring detection

karamanolev:


karamanolev:
This is a POC, so definitely not merge ready. It works well, but fails in a case like that:

    def f(test_arg=some_value[:3]):

with the issue being there's a colon that's not the start-of-function-body colon. That can be worked around by expecting the colon to be right after a closing paren, which is even harder to break, but still possible:

    def f(test_arg=some_value[g():3]):

In order to fix that completely, a parenthesis counter will need to be implemented. Not that difficult, but it will complicate the code even more.

bpmcl:
@karamanolev I can confirm that this approach works in our code base.

karamanolev:
@twolfson I implemented the bracket balancing check mentioned in the issue and it seems to be working. I added the edge cases to the tests and it also seems to be working. @bpmcl can you give the new implementation a spin? WDYT?

bpmcl:
@karamanolev All good here.

twolfson:
I feel a little iffy on this. Maybe to improve our confidence in the code we can break it out into its own file with its own tests?

karamanolev:
@twolfson It had a separate TestCase just for the function anyway, but I've split both the code and the tests in their own separate files and along with that, I slightly improved the code structure now that it's separate.

test/data/doubles_multiline_string.py
```diff
@@ -1,7 +1,7 @@
-""" This "should"
+s = """ This "should"
```
twolfson: Is this considered a docstring by the AST? If not, I'd consider this a regression by our detector =/

test/test_docstring_detection.py
```diff
@@ -0,0 +1,30 @@
...
+        self.assertEqual(self._get_docstring_tokens('data/singles.py'), set())
+        self.assertEqual(self._get_docstring_tokens('data/singles_multiline_string.py'), set())
+        self.assertEqual(self._get_docstring_tokens('data/singles_noqa.py'), set())
+        self.assertEqual(self._get_docstring_tokens('data/singles_wrapped.py'), set())
+
+    def test_get_docstring_tokens(self):
+        with open(get_absolute_path('data/docstring_doubles.py'), 'r') as f:
+            tokens = [Token(t) for t in tokenize.generate_tokens(f.readline)]
+        docstring_tokens = get_docstring_tokens(tokens)
+        self.assertEqual(docstring_tokens, {tokens[1], tokens[18], tokens[45]})
```
twolfson: This assertions is rather vague. Maybe we can adjust it to do a length + value check instead?

```python
self.assertEqual(len(docstring_tokens), 3)
self.assertEqual(docstring_tokens[0].value, 'first docstring')
self.assertEqual(docstring_tokens[1].value, 'second docstring')
self.assertEqual(docstring_tokens[2].value, 'third docstring')

# Alternative setup
self.assertEqual([token.value for token in docstring_tokens], [
  'first docstring',
  'second docstring',
  'third docstring',
])
```

twolfson:
Ah, k. I totally missed that test case on the first go. Thanks for breaking it out. I've left some comments on the code itself. I would love to see even more test scenarios (e.g. indented parameters for a class, default parameters):

```python
class Foo(bar,
                 baz):
    """docstring"""

class Foo(bar="""
    baz"""):
    """docstring"""

if True:
    """This isn't a docstring but has proper indentation pattern"""
```

Also, are there any docs which define when something is considered a docstring and not a docstring (e.g. valid for `class`, `def` but not for `if`, `with`)

test/data/doubles_multiline_string.py
karamanolev: When there is no assignment, it's considered a docstring. When there is an assigment, it's not considered a docstring and I believe that's the correct behavior. I just wanted all existing test cases to not contain docstrings, i.e. make them test existing behavior and add new tests for docstrings. Is there something I'm missing?

twolfson: Ah, you're right. `ast.get_docstring(ast.parse('"""foo"""'))` evaluates  to `'foo'`

twolfson:
For reference it looks like Python identifies the first string of an async function, normal function, class, and module as a docstring:

https://github.com/python/cpython/blob/v3.6.1/Lib/ast.py#L192-L212

karamanolev:
@twolfson I've improved the code as per your requests. As for the docs - this is the relevant PEP AFAIK: https://www.python.org/dev/peps/pep-0257/ and this is a copy-pasted section with the relevant info:

What is a Docstring?
A docstring is a string literal that occurs as the first statement in a module, function, class, or method definition. Such a docstring becomes the __doc__ special attribute of that object.

flake8_quotes/docstring_detection.py
```diff
@@ -0,0 +1,74 @@
...
+    state = STATE_EXPECT_MODULE_DOCSTRING
+    # The number of currently open parentheses, square brackets, etc.
+    # This doesn't check if they're properly balanced, i.e. there isn't ([)], but we shouldn't
+    # need to - if they aren't, it shouldn't parse at all, so we ignore the bracket type
+    bracket_count = 0
+    docstring_tokens = set()
+
+    for token in tokens:
+        if token.type in TOKENS_TO_IGNORE:
+            continue
```
twolfson: It seems like we already ignore tokens if they aren't in the `if/elif` chain. Can we remove this `continue` clause?

test/test_docstring_detection.py
```diff
@@ -0,0 +1,44 @@
...
+from test.test_checks import get_absolute_path
+
+
+class GetDocstringTokensTests(TestCase):
+    def _get_docstring_tokens(self, filename):
+        with open(get_absolute_path(filename), 'r') as f:
+            tokens = [Token(t) for t in tokenize.generate_tokens(f.readline)]
+        return get_docstring_tokens(tokens)
+
+    def test_get_docstring_tokens_absent(self):
```
twolfson: While these tests work, it feels weird to conflate the 2 sets of test files as it will cause extra work between the 2 test suites if we ever want to add a new edge case

We should have a different set of test files for verifying our docstring extraction works as expected; each with a single purpose (e.g. 1 file for classes with many different types of indentation, 1 file for modules (or I guess multiple))

twolfson:
Looking good but there's room for a lot more testing so we have higher confidence =/

flake8_quotes/docstring_detection.py
karamanolev: We can't. We don't ignore tokens if they aren't in the if/elif chain. The last statement is
```        # The token is not one of the recognized types. If we're expecting a colon, then all good,
        # but if we're expecting a docstring, it would no longer be a docstring
        elif state in [STATE_EXPECT_MODULE_DOCSTRING, STATE_EXPECT_CLASS_DOCSTRING,
                       STATE_EXPECT_FUNCTION_DOCSTRING]:
            state = STATE_OTHER```
which catches all the tokens not in the chain and resets the state if we're expecting a docstring. If we didn't ignore the spacing/indentation, then they would also reset the state rendering the code completely broken.

File: test/test_docstring_detection.py (hunk 4 above)
Comment (karamanolev) [inline test/test_docstring_detection.py]: I can separate the files for test_checks and test_docstring_detection, as in the first case we just need one docstring to test quote handling and for the second we need all the cases. I'm a bit skeptical about separating the module, class, function docstring tests, as while they might work in isolation, there could be bugs when all of them are present, which we would miss.

I don't know why you're focusing on indentation. Indentation doesn't matter regarding something being or not being a docstring. Our code ignores indentation. Basically nothing in flake8-quotes's code works around indentation. Why would indentation matter and why would we design tests around something that we really don't care about?

**Comment by karamanolev** [comment]
@twolfson Comments regarding specific issues are above. As for the general "room for more testing" - I've outlined above some points re tests improvement, but I think it's more about the structure/organization, not the content. I've already added all the edge cases that stem from the implementations. Thins like indentation don't matter for the implementation and therefore I think they don't have a place in our tests. The only corner case I could think of is the parens/brackets and colons inside the function parameters, but that's already tested. If you know of something else that's an edge case - please let me know.

File: flake8_quotes/docstring_detection.py (hunk 3 above)
Comment (twolfson) [inline flake8_quotes/docstring_detection.py]: Ah, I see. It might help to have a comment about that in the code. For example:

```python
# If we have a token that doesn't affect our state, then skip it
# DEV: We eagerly skip to prevent hitting accidental resets
if token.type in TOKENS_TO_IGNORE:
```

test/test_docstring_detection.py
twolfson: I think we'll be fine with separating the files -- these should be considered unit tests for now. If we want to find/test more complex cases, then we can add them as more files and search for them via fuzzing.

I'm not focusing on indentation, I'm focused on covering the base cases directly as well as testing any potential edge cases. This is regardless of the current implementation as we want tests to work even if we switch the underlying setup. It just so happens though that Python mostly fluctuates around indentation when altering how functions/classes are defined

twolfson:
@karamanolev While the test cases now cover edge cases like complex indentation, we are lacking simple base cases which can help developers debug quickly and easily (e.g. no parameters, 1 line docstrings)

tony:
can this get rebased and get more tests for it?

karamanolev:
@tony I can rebase it if you're willing to write tests to twolfson's requirements.

tony:
Too much on my plate at the moment

oxitnik:
If @karamanolev don't mind, I'd like to work on this issue. here https://github.com/zheller/flake8-quotes/compare/master...oxitnik:docstring-detection rebased branch with some more tests.
@twolfson, what do you think?

karamanolev:
@oxitnik I'd be more than happy to see this come through.

twolfson:
This has been resolved via #73 and released in `1.0.0`. Thanks for all your work everyone!

PR #60, opened 2017-10-23
avoid flake8 warning

dirk-thomas:
```WARNING:flake8.processor:Plugin requested optional parameter "builtins" but this is not an available parameter.```


twolfson:
Hm, I could have sworn that was a valid parameter at some point. Any idea when `flake8` removed support for that `kwarg`?

dirk-thomas:
This seems to be the change in question: https://github.com/PyCQA/flake8/commit/2d3e277b1e0c0a24c30c611558692f95d14a8470

twolfson:
Thanks for the link! It looks like that was over a year old so I think this is safe to land. Landing shortly :shipit:

twolfson:
This has been merged/released in `0.12.0`. Thanks for the PR!

dirk-thomas:
Thank you for the quick merge & release :+1:

PR #62, opened 2017-12-02
Moved to twine for upload

twolfson:
PyPI has dropped support for uploading via `setup.py upload` due to security concerns. The recommended alternative is to use `twine`:

https://packaging.python.org/tutorials/distributing-packages/#uploading-your-project-to-pypi

In this PR:

- Updated `release.sh` to use `twine`


twolfson:
Landing this for an easier release

PR #63, opened 2017-12-07
Run Travis tests on Python 3.4-3.6

johnthagen:


twolfson:
I'm fine with testing the latest version of Python 3 and maybe 1 other common version but testing every minor feels unnecessary. What's the reasoning here?

johnthagen:
Is this line still significant?

```
# If we are testing Flake8 3.x, then install it
  - if test "$FLAKE8_VERSION" = "3"; then pip install flake8==3; fi
```


I looked like `requirements-dev.txt` already specifies `>=3.3.0`.

johnthagen:
I've seen it as a standard practice since even minor versions of Python can vary things in subtle ways. Python seems to CI very quick and cheaply, so it seemed like a win-win.

I also think that since 3.3 is no longer supported, it should be dropped, but since you had a special case I left it in.

What about 2.7, and 3.4-3.6?

twolfson:
Can you link me to repos that do this as a standard practice? I'm use to people mostly testing major versions and calling it "good enough"

johnthagen:
Here's a couple:

https://travis-ci.org/pyca/cryptography

https://travis-ci.org/paramiko/paramiko

https://travis-ci.org/pypa/pip

twolfson:
That Flake8 line comes from `flake8@3` being in beta:

https://github.com/zheller/flake8-quotes/pull/35

If `flake8@3` is now the norm, then we can prob ditch it and move to `flake8@3` everywhere

johnthagen:
How about we just CI on 2.7, 3.4 and 3.6? I don't have a super strong opinion on it.

twolfson:
Ah, okay. You're right: 2.7, 3.4, 3.5, and 3.6 do seem the be the norm. If we drop update this PR to drop `3.3`, then I'm happy to land it =)

johnthagen:
Should be good to go!

twolfson:
Awesome, thanks! I'm not free at the moment but I'll land this by the end of the weekend

twolfson:
Finally free to do this. Landing it now

twolfson:
This has been released in `0.13.0`. Thanks for the PR and great communication =)

PR #66, opened 2018-01-14
Removal of restrictions of new versions of flake8/tox

themusicgod1:
Why these versions?

PR #67, opened 2018-01-14
preliminary debian packaging, first attempt

themusicgod1:
Since flake8-quotes is required by other software, and since there is an effort to try to get that software into Debian GNU/Linux it would be useful to have the files required to be a debian project synced with this repository. This as a side effect would also help get flake8-quotes (and software depended upon it) into Ubuntu, Mint, Tails, Devuan, and dozens of other distributions both current and future if successful. On the Debian side, this effort is tracked in the following Request for Package (RFP): https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=887115

I have made a fork and added what seems to be the minimum necessary changes to actually be a Debian package, though since I am not actually a DM/DD/maintainer the name will eventually have to be changed by whoever with the actual credentials required to be a maintainer the Debian project chooses to be the maintainer, but it's a first attempt anyway. It does build and allows for flake8-quotes
 to be installed on a Debian GNU/Linux system.

This goes best with #66 - otherwise it's not going to work on Debian/Ubuntu except for versions that it will never be included in, anyway.


twolfson:
What are the benefits of having Debian packaging? Can you point me to other Python projects which include their Debian packaging in the repo itself?

themusicgod1:
tribler off the top of my head https://github.com/tribler/tribler

the benefits:

* Reproducible Builds framework (not as useful with an interpretive language like python as it might be with say C but still important with compiled objects)
* Exposure to new users who could install your program, if accepted into debian/ubuntu.  Ms of them, among them developers who could help you both find and fix bugs if 'quotes' turns out to be something useful to them, which since we know 'quotes' is depended upon by a serious application...maybe there be more.  It's hard to say.  But for sure, it's a step closer to 'quotes' being something that can be depended upon to be used for larger, more mature, more permanent projects.    Debian runs everything from MRIs to national governments.  How much of that will involve your project?  Not much at first.  But over time?

twolfson:
I don't think the demand is high enough for this for us to make it worth building/maintaining. I'm trying to rack my brain around other packages I've seen as well as none of them had built-in Debian packaging; most targeted PyPI (e.g. `requests`, `flake8`, `restructuredtext-lint` (which I maintain))

Specifically my experience for `restructuredtext-lint` was someone else built Debian packaging without needing to ask me for anything

I'm quite doubtful someone would install our standalone plugin or any other package without needing to rebuild the wheel of dependency management which seems counterproductive to PyPI

Closing as `wontfix`

PR #68, opened 2018-02-19
Added a note about escaping quotes in the doc

fpuga:
To avoid surprises to the users I think that it will be a good idea explain in the doc that _wrapped_ quotes are always valid.

I make this PR with a suggestion about the explanation test. Feel free to close the pull and write the message that you think is better.

BTW, great plugin.


twolfson:
Looks good to me. Landing now. Thanks for the PR!

fpuga:
:+1: 

Thanks

PR #71, opened 2018-03-28
requirements-dev: remove tox

blueyed:
Tox config has been removed in 09bab2e, and while it might be worth
bringing it back, there is no point in installing tox in a virtualenv
(`python -m venv .venv`) to run the tests without tox therein.


twolfson:
This has been landed/released in `0.14.1`. Thanks for the PR!

PR #72, opened 2018-03-28
Fix stdin handling: use lines instead of filename

blueyed:
flake8 passes through `--stdin-display-name` for the filename, which is
typically not readable.

Ref: https://github.com/gforcada/flake8-isort/pull/48

### DIFF
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -2,18 +2,6 @@ import optparse
 import tokenize
 import warnings
 
-# Polyfill stdin loading/reading lines
-# https://gitlab.com/pycqa/flake8-polyfill/blob/1.0.1/src/flake8_polyfill/stdin.py#L52-57
-try:
-    from flake8.engine import pep8
-    stdin_get_value = pep8.stdin_get_value
-    readlines = pep8.readlines
-except ImportError:
-    from flake8 import utils
-    import pycodestyle
-    stdin_get_value = utils.stdin_get_value
-    readlines = pycodestyle.readlines
-
 from flake8_quotes.__about__ import __version__
 
 
@@ -54,8 +42,8 @@ class QuoteChecker(object):
     MULTILINE_QUOTES['\'\'\''] = MULTILINE_QUOTES['\'']
     MULTILINE_QUOTES['"""'] = MULTILINE_QUOTES['"']
 
-    def __init__(self, tree, filename='(none)'):
-        self.filename = filename
+    def __init__(self, tree, lines):
+        self.lines = lines
 
     @staticmethod
     def _register_opt(parser, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


blueyed:
This might require to bump/set the minimum flake8 version (in setup.py).
If you still want to support flake8 you should also add it to Travis probably, otherwise more code could be removed.

twolfson:
This is too complex for me to review at a glance. Will review by the end of the weekend. Is there a specific version of `flake8` that must be on and are there any edge cases with `tokenize` that this leads to?

blueyed:
Sure.
Better that you check flake8 requirements - IIRC it is 3.2.1, but that might be just what flake8-isort was using already when I've fixed the same issue there (see the link).

Don't know about edge cases with `tokenize`, but do not understand the question probably.

Instead of using `lines` there might be an API in flake8 to get tokens already, but that would be something for another round.

I will rebase/squash the fixups now, so that it is easier for you to review.
For reference, the current HEAD is bb73a8e.

twolfson:
Revisiting this now...

flake8_quotes/__init__.py
```diff
@@ -119,32 +107,24 @@ def parse_options(cls, options):
...
-            return readlines(self.filename)
-
     def run(self):
-        file_contents = self.get_file_contents()
-
-        noqa_line_numbers = self.get_noqa_lines(file_contents)
-        errors = self.get_quotes_errors(file_contents)
-
-        for error in errors:
+        self._tokens = [Token(t)
```
twolfson: **Required change:** While I understand that we're trying to cache tokens for `run()`, it feels off that we're setting a class property outside of `__init__`. Additionally, this reduces testability by disallowing for passing `tokens` as a parameter to other methods

Requesting to make `tokens` a passable parameter to methods like `self.get_quotes_errors()`:

```python
tokens = [Token(t) ...]
for error in self.get_quotes_errors(tokens)
```

flake8_quotes/__init__.py
```diff
@@ -119,32 +107,24 @@ def parse_options(cls, options):
...
+                        for t in tokenize.generate_tokens(
```
twolfson: **Required discussion:** After doing some research, it looks like `flake8` supports passing tokens instead of lines in exactly the same manner as this. Is there any reason we aren't accepting `tokens`/using `file_tokens` as an `__init__` parameter?

https://github.com/PyCQA/flake8/blob/3.2.1/src/flake8/processor.py#L96-L124

http://flake8.pycqa.org/en/latest/plugin-development/plugin-parameters.html#indicating-desired-data

This has been around for at least 2 years so we're happy to use it

test/test_checks.py
```diff
@@ -35,21 +36,21 @@ class DoublesOptions():
         QuoteChecker.parse_options(DoublesOptions)
 
     def test_multiline_string(self):
-        doubles_checker = QuoteChecker(None, filename=get_absolute_path('data/doubles_multiline_string.py'))
-        self.assertEqual(list(doubles_checker.get_quotes_errors(doubles_checker.get_file_contents())), [
-            {'col': 0, 'line': 1, 'message': 'Q001 Remove bad quotes from multiline string'},
+        doubles_checker = QuoteChecker(None, lines=get_lines('data/doubles_multiline_string.py'))
+        self.assertEqual(list(doubles_checker.run()), [
+            (1, 0, 'Q001 Remove bad quotes from multiline string', QuoteChecker),
```
twolfson: **Required change:** This noise is overkill and makes the test results less clear as to what actually broke. Please revert these and make a method like `get_tokens()` and pass that to the appropriate methods. There's no need to make trouble for ourselves

twolfson:
Closing due to inactivity for almost 1 year

PR #73, opened 2018-04-07
Docstring detection (continue #59)

oxitnik:
Continues @karamanolev 's  pull request #59.


flake8_quotes/__init__.py
```diff
@@ -163,11 +191,22 @@ def get_quotes_errors(self, file_contents):
             #   "foo"[0] * 3 = " * 3 = """
             #   "foo"[0:3] = "fo
             #   """foo"""[0:3] = """
+            is_docstring = token in docstring_tokens
             is_multiline_string = unprefixed_string[0] * 3 == unprefixed_string[0:3]
             start_row, start_col = token.start
 
+            # If our string is a docstring
+            if is_docstring:
```
twolfson: We should prob add a comment about how `is_docstring` **must** come before `is_multiline_string` as docstrings are a subset. I can take care of this when landing the PR though

flake8_quotes/__init__.py
```diff
@@ -163,11 +191,22 @@ def get_quotes_errors(self, file_contents):
...
+                    continue
+
+                yield {
+                    'message': 'Q002 Remove bad quotes from docstring',
+                    'line': start_row,
+                    'col': start_col,
+                }
             # If our string is multiline
-            if is_multiline_string:
+            elif is_multiline_string:
```
twolfson: Note to self: Update the comment above this line to "Otherwise if"

flake8_quotes/docstring_detection.py
```diff
@@ -0,0 +1,74 @@
...
+    for token in tokens:
+        if token.type in TOKENS_TO_IGNORE:
+            continue
+        if token.type == tokenize.STRING:
+            if state in [STATE_EXPECT_MODULE_DOCSTRING, STATE_EXPECT_CLASS_DOCSTRING,
+                         STATE_EXPECT_FUNCTION_DOCSTRING]:
+                docstring_tokens.add(token)
+                state = STATE_OTHER
+        # A class means we'll expect the class token
+        elif token.type == tokenize.NAME and token.string == 'class':
```
twolfson: **Optional change:** It might be a good idea to include an example of the line which would generate this scenario. Similarly for other conditional statements. For example:

```python
# class Foo
```

test/data/docstring_doubles.py
```diff
@@ -0,0 +1,68 @@
+
```
twolfson: **Optional discussion:** Is the leading newline intentional here?

test/test_docstring_detection.py
```diff
@@ -0,0 +1,54 @@
...
+        self.assertEqual(self._get_docstring_tokens('data/singles.py'), set())
+        self.assertEqual(self._get_docstring_tokens('data/singles_multiline_string.py'), set())
+        self.assertEqual(self._get_docstring_tokens('data/singles_noqa.py'), set())
+        self.assertEqual(self._get_docstring_tokens('data/singles_wrapped.py'), set())
+
+    def test_get_docstring_tokens_doubles(self):
+        with open(get_absolute_path('data/docstring_doubles.py'), 'r') as f:
+            tokens = [Token(t) for t in tokenize.generate_tokens(f.readline)]
+        docstring_tokens = {t.string for t in get_docstring_tokens(tokens)}
+        self.assertEqual(docstring_tokens, {
```
twolfson: TIL `{...}` can be used for `set([...]}`

flake8_quotes/__init__.py
```diff
@@ -15,6 +15,7 @@
     readlines = pycodestyle.readlines
```
twolfson: We'll need to update the documentation as well but I can take care of that

test/data/docstring_singles.py
```diff
@@ -0,0 +1,71 @@
...
+
+        for i in range(a):
+            ''' not a docstring'''
+            if a < i:
+                ''' also not a docstring '''
+                pass
+
+    class Nested(foo()[:]): ''' inline docstring '''; pass
+
+def foo():
```
twolfson: **Required change:** While I appreciate the intention of these small tests, they're being globbed with parsing the whole file. Can we relocate it to its own standalone file? Maybe put each of the cases into their own files? (e.g. `docstrings_singles_class`, `docstrings_singles_module`)

oxitnik:
I've put each case into its own file? Not sure if I should split tests.

twolfson:
This has been merged/released in `1.0.0`. Thanks for the PR!

Issue #77, opened 2018-10-03
Add a rule to prevent unnessary \' in single-quoted strings


5j9:
PEP-8 states:

> When a string contains single or double quote characters, however, use the other one to avoid backslashes in the string. It improves readability.

Therefore ``'isn\'t'`` should be rewritten as ``"isn't"``, but `flake8-quotes` currently does not recognize the first form as an error. I'd like to suggest adding it.

twolfson:
I think we're open to an option like that. Maybe call it `nested_quotes: same|different`. Unfortunately I don't have time to build this feature so I'm going to leave this issue open as "help wanted"

JoeHitchen:
I'm willing to have a go tackling this, if we can chat through the requirements.

twolfson:
I've got a bunch of items in the air at the moment, I'll reply with some thoughts by the end of next weekend

JoeHitchen:
The requirements for inline strings seem fairly clear:    (assuming single quote preference)
`'string'` - Good
`'str"ing'` - Good
`'str\'ing'` - Bad (Q0XX?)
`'st"ri\'ng'` - Good (One of them will have to be escaped)
`"string"` - Bad (Q000)
`"str'ing"` - Good
`"str\"ing"` - Bad (Q000)
`"st'ri\"ng"` - Bad (Q000)

Are there any considerations for docstrings/multiline strings? If individual quotes don't need to be escaped in the string body then there is no issue?

My main queries revolve around how best to fit this change into your existing code structure, particular around the tests.

twolfson:
Ah, listing out those test cases helps a bunch. I was projecting personal JS conventions here (i.e. encourage escaping, such that there's an error if we don't -- we don't want the exact opposite of this)

It seems like this is strictly to discourage escaping but prefer the specified quote

Let's go with Q001 for our new error

I concur; I don't think we need to enforce for docstrings/multiline strings -- PEP8 discourages escaping to improve readability, docstrings/multiline don't need escaping

This is another good resource for test cases:

https://github.com/millermedeiros/esformatter-quotes/tree/v1.1.0

Let's go with `avoid-escape: true` (default: `false`) for our new option

JoeHitchen:
I'm going with `Q003 Change outer quotes to avoid escaping inner quotes`, because 0-2 are already used.
Don't we want to enforce this new rule by default, since it is the PEP8 recommendation?

I hope to have something to show by middle of next week.

twolfson:
[...]

PR #78, opened 2019-03-31
Avoid Escaped Quotes

JoeHitchen:
Handling Issue #77 - Added new option `--avoid-escape=true(|false)` to enable checking PEP8 preference for avoiding strings with escaped quotes.

PR likely to contain linting issues.


JoeHitchen:
I'm new to Travis and I'm not seeing those errors locally. Could you translate them for me?

twolfson:
Don't have time to review yet. Travis CI site looks like it's running into
linting errors for the new rule (click on the X, navigate to Travis CI,
click on one of the failing Python versions, scroll to the bottom of the
log)

I think we wrote a `test.sh` which is what Travis CI uses (can be verified
by looking at `.travis.yml`

flake8_quotes/__init__.py
```diff
@@ -145,6 +150,11 @@ def parse_options(cls, options):
         if hasattr(options, 'docstring_quotes') and options.docstring_quotes is not None:
             cls.config.update(cls.DOCSTRING_QUOTES[options.docstring_quotes])
 
+        # If allow-escaped specified, add to config
+        if hasattr(options, 'avoid_escape') and options.avoid_escape is not None:
+            avoid_escape = {'true': True, 'false': False}[options.avoid_escape]
```
twolfson: Does the config parser not support booleans? It looks close to `argparse` under the hood (don't have time fully look at the moment). Maybe we should change it over from `type='choice'` to `type=bool`?

https://docs.python.org/2/library/argparse.html#type

flake8_quotes/__init__.py
```diff
@@ -223,19 +233,38 @@ def get_quotes_errors(self, file_contents):
...
+                #   'This is a "string"'     -> Good
+                #   'This is a \"string\"'   -> Good
+                #   'This is a \'string\''   -> Bad (Q004)  Escaped inner quotes
+                #   '"This" is a \'string\'' -> Good        Changing outer quotes would not avoid escaping
+                #   "This is a string"       -> Bad (Q000)
+                #   "This is a 'string'"     -> Good        Avoids escaped inner quotes
+                #   "This is a \"string\""   -> Bad (Q000)
+                #   "\"This\" is a 'string'" -> Bad (Q000)
+                
+                middle = unprefixed_string[1:-1]
```
twolfson: Recommended variable name: `string_contents`. Middle doesn't make it clear what we're actually picking

flake8_quotes/__init__.py
```diff
@@ -223,19 +233,38 @@ def get_quotes_errors(self, file_contents):
...
+                
+                # Correct outer quote characters
```
twolfson: This comment sound like we're actually correcting the string. Can we update it to a conditional instead?

```python
# If our string is our preferred type, then verify it doesn't only have escaped quotes of the same type
```

test/test_checks.py
```diff
@@ -171,5 +171,61 @@ class Options():
         ])
 
 
+class AvoidEscapeTestChecks(TestCase):
```
twolfson: I feel like a lot of these are redundant and thus going to increase maintenance work =/ Can we consolidate them with existing tests instead? Maybe make the tests be `class SinglesAllowEscaped`/`SinglesAvoidEscaped`/etc

flake8_quotes/__init__.py
```diff
@@ -223,19 +233,38 @@ def get_quotes_errors(self, file_contents):
...
                     continue
-
-                # Output our error
-                yield {
-                    'message': 'Q000 Remove bad quotes',
-                    'line': start_row,
-                    'col': start_col,
-                }
+                
+                # Incorrect outer quote characters
```
twolfson: Same comment update request:

```python
# Otherwise (non-preferred quote type), verify we contain a preferred quote
```

flake8_quotes/__init__.py
```diff
@@ -223,19 +233,38 @@ def get_quotes_errors(self, file_contents):
...
+                if not self.config['good_single'] in middle or self.config['bad_single'] in middle:
```
twolfson: I don't understand the purpose of the `or self.config['bad_single'] in middle:`. Can you explain it?

JoeHitchen: It's for situations where you have both types of quote in a string `The dev said "It's an edge case"` - Since the escape can't be avoided, I said the original preference should stand.

flake8_quotes/__init__.py
JoeHitchen: I can double check, but when I traced the call stack I seemed to get to `optparse` rather than `argparse`, which has pretty restrictive set of types.

https://docs.python.org/3/library/optparse.html#optparse-standard-option-types

test/test_checks.py
JoeHitchen: Sure, I'll have a think about how best to do this :)

JoeHitchen:
Ahh ok, so it's running the latest version of itself against it's own code base. I'll sort out that too then.

flake8_quotes/__init__.py
twolfson: Ah, I see. It looks like we can prob do this coercion via `callback` support and an inline lambda:

https://docs.python.org/3/library/optparse.html#optparse-option-callbacks

This feels like overkill and I'd be surprised if other `flake8` libraries haven't handled booleans yet. It looks like we can do that via `store_true`/`store_false` and options like `--no-avoid-escape`:

https://docs.python.org/3/library/optparse.html#handling-boolean-flag-options

Going to see how/if that would work in our config file and other `flake8` libraries

twolfson: That was a very annoying rabbit hole. It looks like we are indeed stuck with this strategy. What I learned:

- Some libraries look like they support `--no` based options but they fail to properly support them via `config` (e.g. `flake8-isort`)
- Flake8 requires values for any key in config file to show up in parsed data
    - i.e. `no-avoid-escape` won't show up but `no-avoid-escape = False` would (though it would set `avoid-escape = False` which is very confusing)

Sanest solution to avoid string coercion dancing, use simple boolean flags (e.g. `--avoid-escape`, `--no-avoid-escape`), and stay consistent with other libraries:

- Set up both `--avoid-escape` and `--no-avoid-escape` (with `dest: avoid_escape`) options which use `store_true/false` respectively
    - This allows `avoid-escape = True/False` in config files
    - This allows `flake8 --no-avoid-escape`
    - Reference diff on top of PR: https://github.com/zheller/flake8-quotes/compare/zheller:450806c...zheller:5ccd2f0

flake8_quotes/__init__.py
twolfson: I believe that's not a case we'd show an error for, is it?

twolfson: I think the introduction `avoid_escape`/better PEP8 conventions might have flipped good/bad meaning. From my perspective, `good_single` would imply that the string is okay to exist in the middle of a string

I think we have to throw out `good/bad` and move to `same/alt` since good/bad is relative to `avoid_escape`

JoeHitchen:
Refactored the tests - Do they look good now?

I noticed Travis doesn't catch escaped quotes in the application, only in test files - Ignore literally, ignore with noqa, or correct?

twolfson:
Yep, tests look good now

Yea, I noticed that with `test.sh` too. It's caused by us not listing the directory `flake8_quotes`. Ignore it for this PR. I'll fix it when we land this PR

Before I approve this PR, I'd like to finish the discussion around the following open comments (I'll post on them directly so you know which ones)

flake8_quotes/__init__.py
twolfson: Request to resolve before landing

flake8_quotes/__init__.py
twolfson: Request to resolve before landing

flake8_quotes/__init__.py
JoeHitchen: Wrt, the original query, if the preference was for single outer quotes then
> `'The dev said "It\'s an edge case"'`
would be accepted but
> `"The dev said \"It's an edge case\""
would be rejected, with the current behaviour. The line in question is for that rejection case. Happy to change the behaviour there if you'd prefer, it's not specified explicitly in any style guide I've seen.

Wrt, Good/Bad. The current convention makes sense if you see it as referring to the outer quotes, rather than something in the string content. This interpretation also makes sense with `good_multiline`/`bad_multiline` etc where you wouldn't normally expect those values to appear in the body. (For what it's worth, I understood this straight away, but had a much harder time with the semantics elsewhere in the project!)

flake8_quotes/__init__.py
JoeHitchen: Noted, and I will try this out when I next get a chance :)

flake8_quotes/__init__.py
twolfson: Okay, glad to hear it's not a semantics issue. I don't have a lot of time so I'll keep it terse:

From what I can think of, we still don't care about `self.config['bad_single'] in middle`. We're guaranteed that the outer quote is "bad single" due to "good single" always using a `continue` in the previous conditional. The `not self.config['good_single'] in middle` should cover everything

- `"foo"`, bad and doesn't contain good single
- `"f'oo"`, good and does contain good single
- `"The dev said \"edge case\""`, bad and doesn't contain good single
- `"The dev said \"It's an edge case\""`, good and does contain good single

flake8_quotes/__init__.py
JoeHitchen: Your changes looked good. Have squashed, rebased, and pushed.

flake8_quotes/__init__.py
JoeHitchen: The case here would be recommending `"The dev said \"It's an edge case\"" -> 'The dev said "It\'s an edge case"'` to preserve the original preference when an escape is unavoidable. However, there is a strong argument for simply allowing either case. What behaviour would you prefer?

twolfson: The primary purpose of this repo is to ensure quotes stay consistent. Avoiding escaping by switching to alternative quotes is a secondary priority so we should defer to keeping quotes consistent if we need an escape in either case (i.e. require `good_single` be used)

JoeHitchen: I think we're in agreement then, and the code behaves as desired. The first part of the L261 (now L263) `not self.config['good_single'] in string_contents` avoids throwing an error if the string contains a good single. The second part of that line `or self.config['bad_single'] in string_contents` then *does* throw the error in the case being discussed here.

twolfson: Maybe I've over-introduced my own perspective on the goals on the package then. If what you're saying is true, then It feels like we're breaking backwards compatibility with no easy "off" switch which I'd really like to avoid

JoeHitchen: Sorry, had a busy week. I see how this change is also a breaking change - I'll push a change to remove it.

twolfson: Cool, no worries. As a heads up, I'll be leaving for a 1 month vacation in a little over a week so I prob won't land this PR if we don't finish it by Friday April 19 (so there's time for me to revert if need be)
@zheller Might be around to land it but he hasn't dealt with this repo in a while =/

flake8_quotes/__init__.py
```diff
@@ -223,19 +235,38 @@ def get_quotes_errors(self, file_contents):
...
-                if self.config['good_single'] in unprefixed_string:
+                #   'This is a string'       -> Good
+                #   'This is a "string"'     -> Good
+                #   'This is a \"string\"'   -> Good
+                #   'This is a \'string\''   -> Bad (Q004)  Escaped inner quotes
+                #   '"This" is a \'string\'' -> Good        Changing outer quotes would not avoid escaping
+                #   "This is a string"       -> Bad (Q000)
+                #   "This is a 'string'"     -> Good        Avoids escaped inner quotes
+                #   "This is a \"string\""   -> Bad (Q000)
+                #   "\"This\" is a 'string'" -> Bad (Q000)
```
twolfson: I believe this is no longer an error case, I'll correct this before landing

flake8_quotes/__init__.py
```diff
@@ -223,19 +235,38 @@ def get_quotes_errors(self, file_contents):
...
             else:
-                # If our string is a known good string, then ignore it
-                #   (')foo' -> good (continue)
-                #   "it(')s" -> good (continue)
-                #   (")foo" -> possibly bad
-                if self.config['good_single'] in unprefixed_string:
+                #   'This is a string'       -> Good
+                #   'This is a "string"'     -> Good
+                #   'This is a \"string\"'   -> Good
+                #   'This is a \'string\''   -> Bad (Q004)  Escaped inner quotes
```
twolfson: It looks like we say Q004 here but it's actually Q003, I'll correct this before landing (same with other Q004 references)

twolfson:
Ugh, I forgot documentation x_x Adding to release commit =/

twolfson:
This has been released in `2.0.0`. Thanks again for all your help!

JoeHitchen:
Cheers for guiding me through the process, it's been a useful experience!
If anything come up, or you've got something else you think I'd be interested in, let me know :)

PR #88, opened 2019-11-04
Add [flake8] to readme cfg example

cxong:
To make it more obvious how it should be applied


twolfson:
Looks good to me. Landing now. Thanks for the PR!

PR #96, opened 2020-04-18
Added quadruple quote ending support

### LINKED ISSUE #95: Multiline quote raise wrong error
As you wrote:
We follow the PEP8 conventions to avoid backslashes in the string. So, no matter what configuration you are using (single or double quotes) these are always valid strings
```
s = 'double "quotes" wrapped in singles are ignored'
s = "single 'quotes' wrapped in doubles are ignored"
```
Many lines work fine, except some:
```
a = '''"Some text"'''
```
raise error: Q001 Remove bad quotes from multiline string
If I try to replace it:
```
a = """f"Some text""""
```
There python error: SyntaxError: EOL while scanning string literal

So, now it could be only:
```
a = '''"Some text"'''
a = """f"Some text\""""
```

version = flake8-quotes 3.0.0

twolfson:
As reported by @Zoynels in #95, we've got a damned if we do/damned if we don't scenario for quadruple quote endings in strings

We want a string which has contents `foo"` in the end of a multiline string

If we use double quotes, then the Python parser breaks:

```
>>> a = """ "hi""""
  File "<stdin>", line 1
    a = """ "hi""""
                  ^
SyntaxError: EOL while scanning string literal
```

If we use single quotes, then `flake8-quotes` complains as it doesn't see its desired `"""`

```python
a = ''' "hi"'''
#  Q001 Remove bad quotes from multiline string
```

As a result, we've added an explicit exception for this scenario

In this PR:

- Added quadruple quote definitions and exception
- Added tests
- Fixes #95 

@zheller @Zoynels Once one of you reviews/approves, then I'll land this PR

### DIFF
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -42,10 +42,12 @@ class QuoteChecker(object):
     MULTILINE_QUOTES = {
         '\'': {
             'good_multiline': '\'\'\'',
+            'good_multiline_ending': '\'"""',
             'bad_multiline': '"""',
         },
         '"': {
             'good_multiline': '"""',
+            'good_multiline_ending': '"\'\'\'',
             'bad_multiline': '\'\'\'',
         },
     }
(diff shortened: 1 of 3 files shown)


twolfson:
@zheller @Zoynels Would either of you care to review this?

twolfson:
This has been landed/released in `3.1.0`. Thanks for the review =)

PR #99, opened 2020-05-12
Fixed mistaking key access as docstring

### LINKED ISSUE #97: Buggy behavior with empty dict literal and string:
Consider the following example:

```
def test():
    {}["a"]
```

and this configuration:

```
[flake8]
max-line-length = 119
select = Q0
inline-quotes = double
```

this ends up with `Q002 Remove bad quotes from docstring` when there is no docstring here involved.

Splitting this over two lines like:

```
def test():
    d = {}
    d["a"]
```

Resolves it.

twolfson:
As reported in #97, we were mistaking key access on the first line of functions as docstrings. This PR fixes that with a patch via @oxitnik. In this PR:

- Added fix to stop recognizing key access on the first line of functions/modules/classes as docstrings
- Fixes #97

### DIFF
--- a/flake8_quotes/docstring_detection.py
+++ b/flake8_quotes/docstring_detection.py
@@ -63,8 +63,14 @@ def get_docstring_tokens(tokens):
         # Count opening and closing brackets in bracket_count
         elif token.type == tokenize.OP and token.string in ['(', '[', '{']:
             bracket_count += 1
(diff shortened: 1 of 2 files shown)


twolfson:
Considering PR approved as patch was written by @oxitnik and reviewed by me. Landing and releasing now

Issue #82, opened 2019-06-21
strings on multiple lines with different quotes


samuelcolvin:
I'm really keen to use this on [pydantic](https://github.com/samuelcolvin/pydantic), however there's one blocking issue

Take the following example:

```py
        raise ConfigError(
            "validators should be used with fields and keyword arguments, not bare. "
            "E.g. usage should be `@validator('<field_name>', ...)`"
        )
```

I would say both lines should use double quotes so they're the same even though only one line includes single quotes.

I'm aware this would make analysis more complicated, but is it something you would consider? Either as default or via an option.

samuelcolvin:
not a big worry if this isn't possible since I can use `# noqa: Q000` in the few places where this happens.

twolfson:
I think we're open to supporting this but I concur it will add a layer of complexity to everything. One other option aside from `noqa` is to use escaping of the quotes to keep everything consistent:

```python
'validators should be used with fields and keyword arguments, not bare. '
'E.g. usage should be `@validator(\'<field_name>\', ...)`'
```

That being said, I'll leave this issue open as "help wanted" if anyone wants to explore it. Maybe we could call the option `adjacent_line_quotes = False (default)/True`

MaZZly:
I would also be interested in this feature/option

iMakedonsky:
Also interested in this feature

plinss:
I'd like it too, but it needs to allow the quotes to differ if doing so avoids escapes. e.g.

```
        raise ConfigError(
            'validators should be used with "fields" and "keyword" arguments, not bare. '
            "E.g. usage should be `@validator('<field_name>', ...)`"
        )
```

should not be an error.

PR #101, opened 2020-06-04
Added variable docstring support

### LINKED ISSUE #100: Support variable docstrings
While not official, many documentation generators, e.g. Sphinx, pdoc, PyCharm, etc., recognize docstrings after variables.

flake8-quotes however classifies these as multi-line strings, so when using single quotes for multi-line strings and double quotes for docstrings, I get a bunch of false positives.

Example:
```
class Foo:
    """Class docsting."""

    def __init__(self):
        """Constructor docstring."""
        self.var = '''
            multiline string
        '''
        """var docstring."""
```

plinss:
logical lines that begin with a string and don't follow other docstrings are now considered to be docstrings
fixes #100
requires flake8>=3.0
drop support for python 2.7 and 3.4
added some type assertions
update tests for new docstring rules and to call flake8 directly


plinss:
I wasn't able to remove the old docstring logic completely, it's still needed to detect inline docstrings. But a lot of other code was able to go away. 

Given that we now rely on flake8 to determine what a logical line is, the testing code had to change to call flake8 rather than the checker directly. The checker now also returns 1 based column numbers (using the line and column reported by flake8s tokenizer).

Now any logical line with a leading string is considered a docstring, unless if follows another docstring. I did the previous line check to match previous behavior, but I'm not sure if I should have.

A second docstring is not processed as a docstring (AFAICT), but it's still logically a docstring since it's ignored by the python interpreter, so it probably shouldn't be considered a multiline string either as it is now.

I'm thinking to add another error code for duplicate docstrings, but still consider them docstrings. Though checking for duplicate docstrings seems out of scope for this module... I welcome your feedback on that point.

If this PR works for you, I'll add code to fix #82 shortly. I know it's a significant change.

I also added type annotations (mostly) which required dropping support for Python 3.4 as well. This should still work with 3.4 if we just strip the type annotations or switch them all to comment style.

twolfson:
Afk for a while today. Might not be able to review until tomorrow or the
weekend even

At first glance, seems like overreach in the PR =/ If we're just changing
from physical to logical lines and dropping our own parsing support, then
I'd imagine that only the init parameter changes and we drop some
conditional logic

One example of overreach is this looks like it includes changes like adding
a typing convention which I feel is a whole other discussion since it
should require willingness of contributors to maintain and lint that

plinss:
No problem on the timing, it's a significant change in behavior, take your time. I can run on my fork in the interim (or if you decide you don't want to go this route, which is also totally fine).

The type annotations are minor (and just a habit at this point, I find they really do help eliminate bugs), I'm happy to remove them or move them to a separate PR. I don't mean to impose extra workflow or maintenance burden on you. They're really ignored by everything in Python 3.5+ unless you add specific type checks (like mypy or flake8 extensions to require and validate them). If I convert them to comments then they're really ignored by everything except the type check tooling, I just find the comment variant less easy to read and maintain than the annotation. Let me know, it's minutes of work for me.

Aside from that, the only changes not related to the parsing change are dropping Python 2 support; dropping old flake8 support, since it's now dependent on >=3.0 for the logical line processing (the change was simplifying the option processing); and dropping the noqa support, since flake8 handles that for us now too. 

As I suspected the primary quote logic didn't change at all (except adjusting the return types to match what flake8 expects since it's calling the iterator directly now (I just noticed I forgot to return the type, I'll add that))

I did bump the major version as you mentioned due to the flake8 3.0 dependency, I'm not sure if a major version increment is warranted, but totally your call.

The bulk of the diff is the testing code, which was needed since we can't call directly into the checker without computing the logical lines, so we just run flake8 now and parse its output (no sense in trying to replicate that behavior and getting it wrong/out of sync over time). Aside from how the tests are run, the only changes are updating the column numbers, and syncing to the new docstring detection which turns some previously multiline strings into valid docstrings.

flake8_quotes/__about__.py
```diff
@@ -1 +1 @@
-__version__ = '3.2.0'
+__version__ = '4.0.0'
```
twolfson, line 1: We can prob drop this change, I usually do the update via `./release.sh` which adds this change to the `git tag` commit for good measure/traceability

```suggestion
__version__ = '3.2.0'
```

flake8_quotes/__init__.py
```diff
@@ -210,21 +149,16 @@ def get_quotes_errors(self, file_contents):
             #   "foo"[0] * 3 = " * 3 = """
             #   "foo"[0:3] = "fo
             #   """foo"""[0:3] = """
-            is_docstring = token in docstring_tokens
+            is_docstring = (token in self.docstring_tokens)
```
twolfson, line 152: What's the purpose of the parentheses here? It seems inconsistent with the similar logical operators below

```suggestion
            is_docstring = token in self.docstring_tokens
```

flake8_quotes/__init__.py
```diff
@@ -160,37 +122,14 @@ def parse_options(cls, options):
...
-
-    def get_quotes_errors(self, file_contents):
-        tokens = [Token(t) for t in tokenize.generate_tokens(lambda L=iter(file_contents): next(L))]
-        docstring_tokens = get_docstring_tokens(tokens)
-
-        for token in tokens:
+    def __init__(self, logical_line, previous_logical, tokens):
+        self.line = logical_line
+        self.tokens = tokens
+        prev_tokens = tokenize.tokenize(lambda L=iter([previous_logical.encode('utf-8')]): next(L))
```
twolfson, line 128: Not your fault as this seems more/less the same as https://github.com/zheller/flake8-quotes/pull/101/files#diff-d7312ce5f3248fb3a2d6e340b7cf0d22L183

```python
tokens = [Token(t) for t in tokenize.generate_tokens(lambda L=iter(file_contents): next(L))]
```

but this is a damn confusing line, we should prob break it up into a more straightforward function

```suggestion
        # Generate `readline()` matching interface for `tokenize.tokenize` and pass it in
        #   https://docs.python.org/3/library/tokenize.html#tokenize.tokenize
        previous_logical_readline_fn = (lambda L=iter([previous_logical.encode('utf-8')]): next(L))
        prev_tokens = tokenize.tokenize(previous_logical_readline_fn)
```

twolfson:
Also, can you explain how logical lines make it easier to handle variable docstrings as opposed to physical lines?

My current take is that they're not too different and we usually need full file state awareness for the docstring parser to work (i.e. don't just care about previous line but all previous lines in file -- e.g. `class` statement could take a lot of line breaks to finish its closing parens)

flake8_quotes/__about__.py
plinss: Cool. Personally I use a CI job to copy the git tag into the setup.py file and in code I grab the version from pkg_resources.

flake8_quotes/__init__.py
plinss: Sorry, that's just my personal coding style leaking through. When I contribute to other people's code I try to match their style but sometimes habits take over.

FWIW, I tend to wrap pretty much all expressions in parentheses, make it easier to read for me and make it obvious for cases like this when an expression is boolean

flake8_quotes/__init__.py
plinss: Agreed, I was trying to minimize changes.

plinss:
Sure, when flake8 breaks the file up into logical lines, each complete statement is passed as one line (even when it spans multiple physical lines), so any docstring is it's own statement and will the the first significant token on the line. 

The docstring detection code for module docstrings sees variable docstrings as module docstrings when only looking at a single logcial line in isolation. If it weren't for the edge case of inline docstrings, e.g. `class Foo: """docstring"""` we could drop the docstring detection code entirely and just look for lines that start with strings.

There's no need for full file awareness and the only reason to look at the previous line at all is to detect:
```
"""docstring"""
"""not a docstring"""
```
where each docstring is passed in as a separate logical line.

Frankly, I consider the second docstring in that example to still be a docstring, just one that will be ignored by pretty much all documentation processors. I'd rather flag it as a 'bad docstring continuation', but still consider it a docstring. e.g. there's no reason for it to have the multiline string style.

If you agree with that reasoning, we can drop the previous line handling entirely and just let them be docstrings, or use it to surface a different error about continuation docstrings.

This PR does in fact detect all the variable docstrings, as well as other non-standard docstrings, e.g.

```
if something:
    """non-standard docstring"""
```

as evidenced by the test suite. Look for places where I changed """not a docstring""" to """variable docstring""" or """non-standard docstring""".

Python still ignores those non-standard docstrings (as well as docstring continuations), they're just not considered part of the standard documentation.

plinss:
Did some more research and found out that what I was referring to as 'continuation docstrings' are actually called 'additional docstrings' and are part of [PEP258](https://www.python.org/dev/peps/pep-0258/#additional-docstrings). While that PEP was rejected, it's still referred to in [PEP257](https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring) which is still active, and 'additional docstrings' are explicitly mentioned along with 'attribute docstrings'.

So I strongly suggest we consider them docstrings and remove the previous line check entirely.

twolfson:
Was just about to review/process replies but Internet has cut out again
today. Hopefully will be able to go through this tomorrow or Sunday

twolfson:
My internet is back, going to give reviewing a go

Updated the title to better reflect the intention of the PR since logical lines seems more auxiliary based on your comment

twolfson:
Yea, I concur that our goal should be to be in alignment with active PEPs (it doesn't look like there's a finalized state)

flake8_quotes/__init__.py
```diff
@@ -160,37 +122,18 @@ def parse_options(cls, options):
...
-        docstring_tokens = get_docstring_tokens(tokens)
-
-        for token in tokens:
+        # Generate `readline()` matching interface for `tokenize.tokenize` and pass it in
+        #   https://docs.python.org/3/library/tokenize.html#tokenize.tokenize
+        previous_logical_readline_fn = (lambda L=iter([previous_logical.encode('utf-8')]): next(L))
+        prev_tokens = tokenize.tokenize(previous_logical_readline_fn)
+        self.docstring_tokens = get_docstring_tokens(prev_tokens, self.tokens)
 
+    def __iter__(self):
```
twolfson, line 135: I tried to find a reference to this in the `flake8` docs but I can't seem to. Where's this documented and can we add a comment linking to it?

test/test_checks.py
```diff
@@ -1,13 +1,24 @@
-from flake8_quotes import QuoteChecker
 import os
+import re
 import subprocess
 from unittest import TestCase
 
+OUTPUT_REGEX = re.compile(r'.*?:([\d]+):([\d]+): (.*)')
```
twolfson, line 6: I'd really prefer to use `flake8` APIs instead of running the command directly =/ It slows down the tests and makes debugging weirder/harder (e.g. print statements are now in `assert` blocks instead of just `stdout`

Unfortunately, I took a shot but don't have time to dig to find a nice default options for `FileChecker` or `FileProcessor` =/

https://flake8.pycqa.org/en/3.8.2/internal/checker.html#processing-files

twolfson: Also, it's good practice to avoid using `.*` when you can. Better to use not class identifiers to avoid accidentally greedy code:

```suggestion
 OUTPUT_REGEX = re.compile(r'[^:]+:([\d]+):([\d]+): (.*)') 
```

Also can we add an example comment line before so someone can better understand what the regex is doing?

flake8_quotes/__init__.py
plinss: I don't find it referenced in the flake8 docs. Frankly the plugin docs are somewhat lacking. I learned about how logical line plugins work by looking at other examples and reading the flake8 source.

test/test_checks.py
plinss: The `.*?` is a non-greedy match, that change doesn't hurt but is unnecessary as the variants are functionally equivalent. If it's a preferred style thing, have at it, it's your project.

Personally I don't see the value in spending time optimizing unit test code, it's not something users will be dealing with and effort spent there should be balanced against developer pain and the amount of development effort going in to the project.

plinss:
Hi Todd,

I appreciate the effort you put into reviewing this PR, and the work you've done on this project already, but I'm not going to spend the effort breaking this PR up. 

Also, when switching to logical_line handling, you already get the variable docstring handling for free, so they're not really distinct changes. You could in theory drop flake8@2 support first without switching to logical_line handling, but I don't see the value.

I had a need for better docstring handling and rather than just complain, I decided to take care of it myself. As a way of partly paying back the value I have already received from this plugin, I decided to contribute my changes. My needs have been met and I'm using my fork in my projects at this point.

It's your project, do what you want with it. You have my contribution. It's a gift, along with my thanks. Feel free to take my code and do what you want with it, or not. I won't be offended either way, and I don't have a need to be credited by having my commits land directly, so feel free to cherry-pick changes in your own way. I have too many other projects going on right now to spend any more time on this facet of this one. I am happy to continue to answer any questions you may have about my code.

FWIW, if you decide to accept my suggestion of treating additional docstrings as docstrings, you can simply remove all the `previous_line` handling code in the constructor, and in `get_docstring_tokens` remove the `for token in prev_tokens` loop and simply initialize `state` to `STATE_EXPECT_MODULE_DOCSTRING` and you should be set. It's all just additional code to preserve old behavior. You'll also have to update the tests, of course, to match the change in behavior.

twolfson:
Unsure what you mean by "it's your project", open source with broad licensing is the public's project imo -- it just so happens that I'm a maintainer here. Using phrasing like "it's your project" seems like a way to sneak out of responsibility here and in future bug fixes for said relevant code ._.

I similarly don't have time to build out this enhancement fully, hence the "help wanted" tag on the original issue

Closing this PR and making a comment on the original issue for future contributors

Thanks for your time and contributions =)

Issue #102, opened 2021-08-18
Friendlier error message


Julian:
Hi. I was looking to add a linter for string literals to a project of mine and found this one, thanks for writing it.

I find the error message likely a bit too unclear for people however.

Specifically it'll tell someone to "remove bad quotes", when instead it seems clearer to say something of the form "Single quote string literal found, this codebase uses double quoted literals" (or vice versa based on the config).

Are you open to changing it?

twolfson:
Not strongly opposed to it, though I could see it breaking some setups which expect consistent human readable output

Do you mind surveying other parts of `flake8` or plugins to see how they handle different user configurations? (e.g. line length messages, tabs vs spaces, casing)

Julian:
Great! Sure thing.

Here's e.g. [pycodestyle](https://github.com/PyCQA/pycodestyle/blob/aa3417b6a51f5912e32d9c8c879e1b9dd660d5f8/pycodestyle.py#L319-L320) which for line length too long will say "(137 > whatyouconfiguredasmaxlinelength" (on the same E501 rule)

And a plugin for checking tabs and spaces [which does the same seemingly](https://gitlab.com/ntninja/flake8-tabs/-/blob/e730ad4b2c82fa2e2af8b5478902efc137abb90d/flake8_tabs/indentationck.py#L79-88), using a single rule but modifying the text for human readability depending on if tabs or spaces are configured.

twolfson:
Awesome, thanks @Julian =D

I'm now much more comfortable with changing the text =) We're prob fine to leave them using the same code. If people eventually request it, then we can make that a breaking change (major release) then.

Do you mind opening a PR to introduce the change? Best approach will prob be putting the full message into the `INLINE_QUOTES`/`MULTILINE_QUOTES` config -- this will futureproof us for i18n if that ever comes up (e.g. syntax/ordering in 1 language can be unpredictable)

https://github.com/zheller/flake8-quotes/blob/461254935334d3877c7c0990a39fe6c74494a207/flake8_quotes/__init__.py#L25-L53

Julian:
Sure thing. Sent.

twolfson:
This has been released in `flake8-quotes@3.3.0`

PR #103, opened 2021-08-20
Make the error message show what kind of string is bad.

Julian:
Closes: #102

### DIFF
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -73,6 +73,8 @@ class QuoteChecker(object):
     DOCSTRING_QUOTES["'''"] = DOCSTRING_QUOTES["'"]
     DOCSTRING_QUOTES['"""'] = DOCSTRING_QUOTES['"']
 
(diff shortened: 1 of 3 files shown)


twolfson (commented):
Ehh, I have a lot of nitpicks around this implementation =/ Will just commandeer rather than try to force mindreading =/ Will update + land by Fri Aug 27

- Would prefer messages be written out as-is inside the `INLINE` and `MULTILINE` configs (good for i18n flexibility)
- New messages are *very* verbose, worry about being too noisy for console feedback
- Disliking how tests are using variables in assertion expectationss (should almost always be avoided as it can lead to more levels of indirection which makes testing unclear)

Julian:
Sure, go for it, thanks.

twolfson:
Looking into this now. As a heads up, this branch is currently failing `./test.sh`. I'll look into why as part of the resolution

twolfson:
That was kind of tedious but it's all back to functional now. Landing PR

twolfson:
This has been released in `flake8-quotes@3.3.0`. Thanks for your work!

PR #104, opened 2021-10-14
Use assertRegex instead of assertRegexpMatches for Python 3.11 compatibility.

tirkarthi:
The deprecated aliases were removed in Python 3.11 in python/cpython#28268 . Fixes below warnings that are error now.

```
/root/checked_repos_clone_1100_1200/flake8-quotes/test/test_checks.py:25: DeprecationWarning: Please use assertRegex instead.
  self.assertRegexpMatches(
/root/checked_repos_clone_1100_1200/flake8-quotes/test/test_checks.py:28: DeprecationWarning: Please use assertRegex instead.
  self.assertRegexpMatches(
/root/checked_repos_clone_1100_1200/flake8-quotes/test/test_checks.py:31: DeprecationWarning: Please use assertRegex instead.
  self.assertRegexpMatches(
```

### DIFF
--- a/test/test_checks.py
+++ b/test/test_checks.py
@@ -11,6 +11,10 @@ class TestChecks(TestCase):
 
 
 class TestFlake8Stdin(TestCase):
+    def setUp(self):
+        if not hasattr(self, 'assertRegex'):
+            self.assertRegex = self.assertRegexpMatches
+
     def test_stdin(self):
         """Test using stdin."""
         filepath = get_absolute_path('data/doubles.py')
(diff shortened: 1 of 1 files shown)


test/test_checks.py
```diff
@@ -11,6 +11,10 @@ def test_get_noqa_lines(self):
 
 
 class TestFlake8Stdin(TestCase):
+    def setUp(self):
+        if not hasattr(self, 'assertRegex'):
+            self.assertRegex = self.assertRegexpMatches
+
```
twolfson, line 17: According to your PR, `assertRegexpMatches` has been deprecated since Python@3.2, which has been out since 2011. No need to worry about Python@2. We can safely drop this fallback =)

https://github.com/python/cpython/pull/28268/files#diff-14c3c1d897d1e85767303f24626acaa6d1ea1162219c72caa240ef9f533ebe99L1636-L1638

https://www.python.org/download/releases/3.2/

```suggestion
```

twolfson (approved):
LGTM

twolfson:
This is great! Thanks for the PR! =D

tirkarthi:
Thanks for the review. The project had Python 2.7 classifier in setup.py. So I thought Python 2 is supported and added fallback. I guess it's good to drop the classifier and also add python_requires https://packaging.python.org/guides/dropping-older-python-versions/

twolfson:
Good call on `python_requires`, it doesn't hard block which is nice =) Dropping the classifier also sounds good 👍 

I don't have capacity to pick this up at the moment but I should be able to do that by the end of next weekend

twolfson:
I've updated the classifiers and release `3.3.1`. Thanks for all your work @tirkarthi!

PR #106, opened 2022-07-14
Update README.rst

pbelskiy:
Thanks for awesome package!

But it's not obvious which quotes are default for doctoring, so experimentally it's double, so I update your readme.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -50,7 +50,10 @@ Q003 Change outer quotes to avoid escaping inner quotes
 Configuration
 -------------
 
-By default, we expect single quotes (') and look for unwanted double quotes ("). To expect double quotes (") and find unwanted single quotes ('), use the CLI option:
+By default, we expect single quotes (') and look for unwanted double quotes ("), and
+expect double quotes (") in docstring. 
+
+To expect double quotes (") and find unwanted single quotes ('), use the CLI option:
 
 .. code:: shell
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -50,7 +50,10 @@ Q003 Change outer quotes to avoid escaping inner quotes
 Configuration
 -------------
 
-By default, we expect single quotes (') and look for unwanted double quotes ("). To expect double quotes (") and find unwanted single quotes ('), use the CLI option:
+By default, we expect single quotes (') and look for unwanted double quotes ("), and
```
twolfson, line 53: I appreciate the effort here! I think this is a little more confusing though, since we're starting to talk about expectations in 2 different places.

It'd prob be better to split up each of the code samples with their relevant default and override settings (e.g. multiline, docstring).

Or simply add a default explainer to the existing comment lines

```
# We also support configuration for multiline quotes (default: """)

# We also support docstring quotes similarly (default: """)
```

What are your thoughts?

> For personal reference: https://github.com/zheller/flake8-quotes/blob/3.3.1/flake8_quotes/__init__.py#L137-L139

pbelskiy: So, we can go deeper here, but I think it's quite enough now just to write default behaviour, it will solve my and other developers misunderstanding.

PR #107, opened 2022-11-02
Replace setup.py with pyproject.toml

skirpichev:
FYI: I see "DEPRECATION: flake8-quotes is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change.", hence this PR.

Second commit adds a simple GA workflow, see an [example](https://github.com/skirpichev/flake8-quotes/actions/runs/3374580149).

Unfortunately, the pyproject.toml support will require setuptools>=61, that's why requires-python option was changed.


skirpichev:
Ok, I assume this is not helpful.

PR #118, opened 2023-11-11
Feature/f string checks

arnimarj:
This PR adds preliminary support for generalized f-strings, as defined in PEP701 (https://peps.python.org/pep-0701/).

With the advent of PEP701, the token stream for a given f-string is richer than before. Take for example the snippet

```python
f"{d.strftime('%d%m')}"
```

In Python 3.11 we'd simply get 

```shell
TokenInfo(type=3 (STRING), string='f"{d.strftime(\'%d%m\')}"', start=(1, 0), end=(1, 23), line='f"{d.strftime(\'%d%m\')}"\n')
TokenInfo(type=4 (NEWLINE), string='\n', start=(1, 23), end=(1, 24), line='f"{d.strftime(\'%d%m\')}"\n')
TokenInfo(type=0 (ENDMARKER), string='', start=(2, 0), end=(2, 0), line='')
```

But in Python 3.12 we get:

```shell
TokenInfo(type=61 (FSTRING_START), string='f"', ...)
TokenInfo(type=55 (OP), string='{', ...)
TokenInfo(type=1 (NAME), string='d', ...)
TokenInfo(type=55 (OP), string='.', ...)
TokenInfo(type=1 (NAME), string='strftime', ...)
TokenInfo(type=55 (OP), string='(', ...)
TokenInfo(type=3 (STRING), string="'%d%m'", ...)
TokenInfo(type=55 (OP), string=')', ...)
TokenInfo(type=55 (OP), string='}', ...)
TokenInfo(type=63 (FSTRING_END), string='"', ...)
TokenInfo(type=4 (NEWLINE), string='\n', ...)
TokenInfo(type=0 (ENDMARKER), string='', ...)
```

If we pass that token stream (where we check `FSTRING_START` as well as `STRING`), we get false-positives, since the double quotes in `'f"'` aren't negated by the single quotes inside the larger string, since the larger string is never seen, since it is no longer a single token.

In order to make this version backwards compatible I stitch together the token stream which comprised a single f-string, and process it as a whole. This should mimic the current semantics.

For codebases which need to support versions older than 3.11 this is probably good behaviour. Other codebases could set the new configuration variable `check-inside-f-strings` which will instead process the original token stream (treating `FSTRING_START` the same as `STRING`). This allows users to have consistent string formatting inside f-strings, e.g.:  `f'{d.strftime('%d%m')}'`, which is a syntax error in Python 3.11 and older.


arnimarj:
When running under Python 3.12 there are regressions, when running in single-quote mode:

```python
f"{a}{b}{d.strftime('%d%m')}"
```

...gave a negative under 3.11, but a positive in 3.12

ThiefMaster:
Ah, glad to see someone is already working on it. I was just about to start but I'll leave it to you then :)

ThiefMaster:
If it's not straightforward to properly handle it, maybe a simple case of using a new error id (e.g. `Q004`) for anything inside f-strings would be a good alternative? That way people can simply disable it if they do not want to be bothered about quote styles inside f-strings

arnimarj:
@JoeHitchen feel free to poke holes in this

JoeHitchen:
Looked good to me, and passed the tests with the Github Actions I've just set up 👍 

Will try to look at what is needed for a release (v3.4.0) next weekend

JoeHitchen:
This has finally been released as 3.4.0

PR #120, opened 2024-02-15
Fix #119

### LINKED ISSUE #119: Incorrect Q003 error
Since 3.4.0 came out I am getting Q003 errors on code such as:

```
string = f'My id is {some_dictionary['id']}'
```

As per f-string changes in python 3.12 it is possible to now use the any quotes within the f-string expression without escaping first.
This code was not flagging before 3.4.0.
It really shouldn't be flagged given that there are no escaped inner quotes here.

arnimarj:
Testbed to investigate https://github.com/zheller/flake8-quotes/issues/119

### DIFF
--- a/flake8_quotes/__init__.py
+++ b/flake8_quotes/__init__.py
@@ -256,14 +256,11 @@ class QuoteChecker(object):
                     yield from self._check_string(token_string, fstring_start, is_docstring)
                     continue
 
-            # otherwise, we check nested strings and f-strings, we don't
-            # check FSTRING_END since it should be legal if tokenize.FSTRING_START succeeded
-            if token.type in (tokenize.STRING, tokenize.FSTRING_START,):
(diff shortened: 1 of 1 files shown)


jnrbsn:
What's holding this back from being merged?

arnimarj:
> What's holding this back from being merged?

I never got around adding proper tests for this, but this does seem to fix the reported issue.

I'll move it out of draft to get comments on it.

jnrbsn:
Who needs to give feedback on this to move it forward? @JoeHitchen maybe?

KristianTashkov:
One note on this fix is that it seems something that used to produce an error before, now doesn't:
```
print(f"Mood is {moods['sunny']}")
```

This doesn't require you to move to uniform choice of quotes in the fstring and in the interpolation string.

